PR #264, opened 2013-09-02
Allow set default cusom repositoty class option

thestanislav:


src/DoctrineORMModule/Options/Configuration.php
```diff
@@ -462,4 +470,27 @@ public function getNamingStrategy()
...
+     * Sets default repository class.
+     *
+     * @param string $className
+     *
+     * @return void
+     *
+     */
+    public function setDefaultRepositoryClassName($className)
+    {
+        $this->defaultRepositoryClassName = $className;
```
bakura10: Add a cast to string.

bakura10:
Replaced by #375, thanks!

PR #268, opened 2013-09-26
enable quoteStrategy-Configuration

Blaimi:
because https://github.com/doctrine/doctrine2/pull/800 will not happen …


Blaimi:
This is a duplicate of #453 #

Issue #295, opened 2013-12-17
Add support for second level caching


bakura10:
Hi,

Doctrine 2.5 will ship with second region caching. Would be good to have it for DoctrineORMModule 1.0: https://github.com/doctrine/doctrine2/blob/master/docs/en/reference/second-level-cache.rst

PR #318, opened 2014-04-20
Add support for second level cache

bakura10:
Addresses #295

Tests need to be written.


bakura10:
Tests and doc have been added @Ocramius. That's now mergeable :).

bakura10:
Again, tests are failing on PHP 5.3 only because of the minimum dependency required in Doctrine 2.5.

stof:
@Ocramius why not using a version check to be compatible with both 2.4 and 2.5, as done in DoctrineBundle ? Forcing ZF users to wait until 2.5 is stable before trying the feature is bad. We want them to be able to test features during the beta and RC phases (otherwise these phases are less useful)

Ocramius:
> why not using a version check

Could be done, I just don't have time to deal with it myself right now.

stof:
@bakura10 can you add a version check in your code to allow keeping the compatibility with the ORM 2.4 ?

bakura10:
Done @stof 

@Ocramius , I think we can merge this and #374 , then upgrading minimum PHP version to PHP 5.4 immediately?

composer.json
```diff
@@ -34,8 +34,8 @@
     "require":           {
         "php":                               ">=5.3.23",
         "doctrine/doctrine-module":          "0.9.*",
-        "doctrine/orm":                      ">=2.4,<2.6-dev",
-        "doctrine/dbal":                     ">=2.4,<2.6-dev",
+        "doctrine/orm":                      ">=2.5,<2.7-dev",
+        "doctrine/dbal":                     ">=2.5,<2.7-dev",
```
stof: The goal of adding a version check is to preserve the compatibility with 2.4 too

bakura10:
Done @stof . @Ocramius , ready to merge?

Ocramius:
@bakura10 needs rebase

bakura10:
This is mergeable @Ocramius !

PR #363, opened 2015-01-04
Prevent Zend\Form\Element\File types inherit of StringLength validator...

mpalourdio:
...when a length is provided
See https://github.com/zendframework/zf2/issues/7096

With annotations, when a file type is specified, the ElementAnnotationsListener attach a string length validator by default. This PR prevents this behavior.


Ocramius:
@mpalourdio thanks, merged!

PR #374, opened 2015-01-21
Use `ResolveTargetEntityListener` as an event subscriber when supported

### LINKED ISSUE #370: Support resolver of interfaces
With the recent addition in Doctrine ORM 2.5 for resolving interface names everywhere, DoctrineORMModule needs to be updated:

https://github.com/doctrine/DoctrineORMModule/blob/1d5b0fdda6200836b0d1dcf637d36e35ea48ad6c/src/DoctrineORMModule/Service/EntityResolverFactory.php#L45

This line has to be changed to 

```php
$eventManager->addEventSubscriber($targetEntityListener);
```

I'm not sure if this is possible to do it without breaking BC until Doctrine 2.5 is out.

bakura10:
Fix #370

The minimum stability has been set to alpha until Doctrine 2.5 is officially released (I think we'll be able to do that once tagging).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,10 +33,10 @@
     ],
     "minimum-stability": "alpha",
     "require":           {
-        "php":                               ">=5.3.23",
+        "php":                               ">=5.4",
         "doctrine/doctrine-module":          "0.9.*",
-        "doctrine/orm":                      ">=2.5,<2.7-dev",
-        "doctrine/dbal":                     ">=2.5,<2.7-dev",
+        "doctrine/orm":                      ">=2.4,<2.7-dev",
+        "doctrine/dbal":                     ">=2.4,<2.7-dev",
         "zendframework/zend-stdlib":         "~2.3",
         "zendframework/zend-mvc":            "~2.3",
         "zendframework/zend-servicemanager": "~2.3",
--- a/composer.json
+++ b/composer.json
@@ -31,11 +31,12 @@
             "email": "guilhermeblanco@hotmail.com"
         }
     ],
+    "minimum-stability": "alpha",
     "require":           {
         "php":                               ">=5.3.23",
         "doctrine/doctrine-module":          "0.9.*",
-        "doctrine/orm":                      ">=2.4,<2.6-dev",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -31,11 +31,12 @@
             "email": "guilhermeblanco@hotmail.com"
         }
     ],
+    "minimum-stability": "alpha",
     "require":           {
         "php":                               ">=5.3.23",
         "doctrine/doctrine-module":          "0.9.*",
-        "doctrine/orm":                      ">=2.4,<2.6-dev",
-        "doctrine/dbal":                     ">=2.4,<2.6-dev",
+        "doctrine/orm":                      ">=2.5,<2.7-dev",
```
danizord: Could be `~2.5.0` ?

bakura10:
Tests are failing becuase 2.5 requires PHP 5.4. I think we can change that when tagging.

danizord:
@bakura10 Can't you change the PHP version here in this PR?

bakura10:
I could but I think we'll do that on merge. I'd prefer not to merge anything related to 2.5 until 2.5 is out, as I know some people are using dev-master branch of DoctrineORMModule.

bakura10:
@stof : I've added the check here too

bakura10:
This one does not need rebase @Ocramius :)

src/DoctrineORMModule/Service/EntityResolverFactory.php
```diff
@@ -42,7 +43,11 @@ public function createService(ServiceLocatorInterface $serviceLocator)
             $targetEntityListener->addResolveTargetEntity($oldEntity, $newEntity, array());
         }
 
-        $eventManager->addEventListener(Events::loadClassMetadata, $targetEntityListener);
+        if (Version::compare('2.5.0') >= 0) {
```
Ocramius: @bakura10 instead of this, why not checking if `$targetEntityListener` implements the subscriber interface?

bakura10:
Done @Ocramius

Issue #353, opened 2014-10-30
Missing Symfony YAML component to dependencies


jbenoit2011:
I tried to generate entities from a YAML file in my ZF2 app, but it failed because Symfony YAML component is not installed. Whereas I did a `composer require doctrine/doctrine-module` to install DoctrineORMModule.

The dependency is not listed on [Packagist](https://packagist.org/packages/doctrine/doctrine-module)
Maybe it should be added to dependencies ?

Here is the call stack:
```shell
user@host:~/sampleapp/server$ php public/index.php orm:generate-entities module/Application/src/Application/Entity
PHP Fatal error:  Class 'Symfony\Component\Yaml\Yaml' not found in /home/user/sampleapp/server/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/Driver/YamlDriver.php on line 712
PHP Stack trace:
PHP   1. {main}() /home/user/sampleapp/server/public/index.php:0
PHP   2. Zend\Mvc\Application->run() /home/user/sampleapp/server/public/index.php:17
PHP   3. Zend\EventManager\EventManager->trigger() /home/user/sampleapp/server/vendor/zendframework/zendframework/library/Zend/Mvc/Application.php:313
PHP   4. Zend\EventManager\EventManager->triggerListeners() /home/user/sampleapp/server/vendor/zendframework/zendframework/library/Zend/EventManager/EventManager.php:207
PHP   5. call_user_func() /home/user/sampleapp/server/vendor/zendframework/zendframework/library/Zend/EventManager/EventManager.php:468
PHP   6. Zend\Mvc\DispatchListener->onDispatch() /home/user/sampleapp/server/vendor/zendframework/zendframework/library/Zend/EventManager/EventManager.php:468
PHP   7. Zend\Mvc\Controller\AbstractController->dispatch() /home/user/sampleapp/server/vendor/zendframework/zendframework/library/Zend/Mvc/DispatchListener.php:114
PHP   8. Zend\EventManager\EventManager->trigger() /home/user/sampleapp/server/vendor/zendframework/zendframework/library/Zend/Mvc/Controller/AbstractController.php:117
PHP   9. Zend\EventManager\EventManager->triggerListeners() /home/user/sampleapp/server/vendor/zendframework/zendframework/library/Zend/EventManager/EventManager.php:207
PHP  10. call_user_func() /home/user/sampleapp/server/vendor/zendframework/zendframework/library/Zend/EventManager/EventManager.php:468
[...]

PR #379, opened 2015-01-28
#353 Add Symfony/Yaml dependency in composer.json

MadCat34:


Ocramius:
@MadCat34 we are not using `Yaml` directly: it's an ORM [optional dependency](https://github.com/doctrine/doctrine2/blob/17a23ea82595d49995f347bf6dae7e218e98f1f4/composer.json#L31).

Since we don't reference it in this package, it should not be included.

MadCat34:
Ok, I hadn't seen it was a suggest for ORM ... :)

PR #388, opened 2015-02-26
update comment block in Module.php as no Module::getAutoloaderConfig() 

samsonasik:
means it must be using composer

### DIFF
--- a/Module.php
+++ b/Module.php
@@ -1,7 +1,6 @@
 <?php
 /**
  * This file is placed here for compatibility with ZendFramework 2's ModuleManager.
- * It allows usage of this module even without composer.
  * The original Module.php is in 'src/DoctrineORMModule' in order to respect PSR-0
  */
 require_once __DIR__ . '/src/DoctrineORMModule/Module.php';


Ocramius:
@samsonasik thanks!

samsonasik:
Actually, after I re-think about it, the comment block itself need to be updated to make it work without composer with Zend\Loader ?

Warm regards,

Abdul Malik Ikhsan

Pada 27 Feb 2015, pukul 05.23, Marco Pivetta <notifications@github.com> menulis:

samsonasik:
or another option. revert this change and add Module::getAutoloaderConfig() ?

Warm regards,

Abdul Malik Ikhsan

Pada 27 Feb 2015, pukul 05.23, Marco Pivetta <notifications@github.com> menulis:

Ocramius:
@samsonasik the module does not need to work without composer

samsonasik:
@Ocramius so this file can be deleted, then ?

mpalourdio:
If you delete it, you need to add a 'classmap' entry in composer.json, that reference the Module class as the ModuleManager expects to find Module.php under /src

Ocramius:
@mpalourdio no need: the ModuleManager only autoloads what isn't autoloadable.

PR #389, opened 2015-03-02
Delete Module.php in root directory

samsonasik:
I hope I don't missunderstand of https://github.com/doctrine/DoctrineORMModule/pull/388#issuecomment-76683250 As if there is no Module::getAutoloaderConfig() , we can't work without composer /cc @Ocramius ;)


Ocramius:
:+1:

PR #390, opened 2015-03-02
travis: PHP 5.6, 7.0 nightly added, 5.3 dropped

TomasVotruba:
To be compatible with [version in composer.json](https://github.com/doctrine/DoctrineORMModule/blob/51c64bfc5a87148379a03139e0c986c09c81d6af/composer.json#L36)

Bash syntax cleanup

Next idea: running coverage only once?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,15 @@
 language: php
 
 php:
-  - 5.3
   - 5.4
   - 5.5
+  - 5.6
+  - 7.0
   - hhvm
 
 matrix:
   allow_failures:
+    - php: 7.0
     - php: hhvm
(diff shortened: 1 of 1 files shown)


Ocramius:
@TomasVotruba thanks! Not sure if speeding up coverage is needed for this repo btw

TomasVotruba:
@Ocramius Thank you.

For me, running coverage is more about best practice then speed. No added value.

Also it's easier to add new matrix run to travis (e.g. hhvm-nightly ATM), when it's not burdened by coverage. Also more convenient for me while sending PR and waiting for tests (faster = more motivated to check).

PR #392, opened 2015-03-12
Use-case: caching module's configuration

Gamempire:
Avoid troubles in case of caching module's configuration

### DIFF
--- a/docs/cache.md
+++ b/docs/cache.md
@@ -47,6 +47,41 @@ In this case you have to specify a custom factory in your `service_manager` conf
 ),
 ```
 
+Please be aware that you can't use *Closures* inside yout module configuration if you want to cache it. To avoid this problem, you have to use a Factory:
+```php
+// module.config.php
+'service_manager' => array(
+    'factories' => array(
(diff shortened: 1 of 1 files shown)


gianarb:
Thanks for this improvement!

Gamempire:
:+1:

docs/cache.md
```diff
@@ -47,8 +47,43 @@ In this case you have to specify a custom factory in your `service_manager` conf
 ),
 ```
 
+Please be aware that you can't use *Closures* inside yout module configuration if you want to cache it. To avoid this problem, you have to use a Factory:
+```php
+// module.config.php
+'service_manager' => array(
+    'factories' => array(
+        'my_memcached_alias' => __NAMESPACE__ . '\Cache\MemcachedFactory'
```
danizord: Config files usually do not have namespaces. Can you replace `__NAMESPACE__` with `YourModule`?

danizord: Ah nvm, I did not saw that it was merged :D

PR #398, opened 2015-04-21
Composer - update for stable version

gianarb:
# Stable dependencies

I'm working to push out of door the first stable release of this module.
In my opinion this is a good starting point to work with stable dependencies.
## Zend Framework 2.3 and 2.4

In this moment we will support this two version of Zend Framework. But we work in dev-develop during developer env becuase we look the future :smile: 
## Minimum stabiliy

Bye `alpha` and `develop` this module will be stable :+1:

PR #399, opened 2015-04-22
Doctrine\Migrations require dialog helper

gianarb:
Reference [#398-issuecomment-95072327](https://github.com/doctrine/DoctrineORMModule/pull/398#issuecomment-95072327) 
## Problem

In this moment we don't manage a dialog helper for doctrine\migrations component.
This Helper is deprecated but doctrine/migrations use it now.

ping @lastzero for feedback :)

### DIFF
--- a/src/DoctrineORMModule/Module.php
+++ b/src/DoctrineORMModule/Module.php
@@ -29,6 +29,7 @@ use Symfony\Component\Console\Helper\QuestionHelper;
 use Doctrine\DBAL\Tools\Console\Helper\ConnectionHelper;
 use Doctrine\ORM\Tools\Console\Helper\EntityManagerHelper;
 use Zend\Stdlib\ArrayUtils;
+use Symfony\Component\Console\Helper\DialogHelper;
 
 /**
  * Base module for Doctrine ORM.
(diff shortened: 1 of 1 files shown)


src/DoctrineORMModule/Module.php
```diff
@@ -142,5 +143,8 @@ public function initializeConsole(EventInterface $event)
         $helperSet->set(new QuestionHelper(), 'question');
         $helperSet->set(new ConnectionHelper($entityManager->getConnection()), 'db');
         $helperSet->set(new EntityManagerHelper($entityManager), 'em');
+        if (class_exists('Doctrine\\DBAL\\Migrations\\Version')) {
```
Ocramius: Why is the check build like this? Can you add a comment?

src/DoctrineORMModule/Module.php
```diff
@@ -142,5 +143,8 @@ public function initializeConsole(EventInterface $event)
...
+            $helperSet->set(new DialogHelper(), 'dialog');
```
Ocramius: Should be done in whatever code adds the migrations stuff, no?

gianarb: I don't know.. For me this is a good idea.. But I can move this dependency how required
https://github.com/doctrine/DoctrineORMModule/blob/master/composer.json#L49

:) Now is require dev.. For me it's a good stuff

gianarb:
Reference #401

PR #405, opened 2015-05-20
Provide the dialog and question helper

BreiteSeite:
Provide dialog helper for symfony/console 2.x and question helper for symfony/console 3.x

Needed (for example) for [doctrine/migrations PR #253](https://github.com/doctrine/migrations/pull/253)

### DIFF
--- a/src/DoctrineORMModule/Module.php
+++ b/src/DoctrineORMModule/Module.php
@@ -19,6 +19,8 @@
 
 namespace DoctrineORMModule;
 
+use Symfony\Component\Console\Helper\DialogHelper;
+use Symfony\Component\Console\Helper\QuestionHelper;
 use Zend\ModuleManager\Feature\ControllerProviderInterface;
 use Zend\ModuleManager\Feature\ConfigProviderInterface;
 use Zend\ModuleManager\Feature\InitProviderInterface;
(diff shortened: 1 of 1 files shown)


src/DoctrineORMModule/Module.php
```diff
@@ -139,7 +138,14 @@ public function initializeConsole(EventInterface $event)
         $entityManager = $serviceLocator->get('doctrine.entitymanager.orm_default');
         $helperSet     = $cli->getHelperSet();
 
-        $helperSet->set(new QuestionHelper(), 'question');
+        if (class_exists('Symfony\Component\Console\Helper\QuestionHelper')) {
+            $helperSet->set(new \Symfony\Component\Console\Helper\QuestionHelper(), 'question');
+        }
+
+        if (class_exists('Symfony\Component\Console\Helper\DialogHelper')) {
+            $helperSet->set(new \Symfony\Component\Console\Helper\DialogHelper(), 'dialog');
```
stof: you should set the DialogHelper only when the QuestionHelper is not available, otherwise you will still trigger deprecation notices here

BreiteSeite: I set both to keep backwards compatibility for [symfony/console ~2.5](https://github.com/symfony/Console/tree/2.5/Helper) where both are available.

gianarb:
Sorry it is Double #399

src/DoctrineORMModule/Module.php
stof: The issue is that the instantiation here will trigger the deprecation warning

BreiteSeite: Ah.. i see this was added in symfony/console ~2.7. I could call the constructor with false to avoid the deprecation warning, but this would be an error in PHP 7 with symfony/console < 2.7, because of the strict argument count check added in PHP 7.

@stof Do you think this is an issue? If yes i would need to check with ReflectionFunctionAbstract::getNumberOfParameters or similar for the signature of the constructor. If no, i would just change the code to new DialogHelper(false)...

gianarb:
Thanks! 
Merged bde0b1b

PR #413, opened 2015-05-29
Fixed missing version

gianarb:
I have forgotten to change version of this lib

:sos: 

ping @Ocramius

### DIFF
--- a/src/DoctrineORMModule/Version.php
+++ b/src/DoctrineORMModule/Version.php
@@ -29,5 +29,5 @@ namespace DoctrineORMModule;
  */
 class Version
 {
-    const VERSION = '0.8.0';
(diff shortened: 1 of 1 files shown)


mpalourdio:
Will you re-tag 0.9.0 ? It may break those that are bases on composer.lock no ? You'd better tag 0.9.1 IMO.

Ocramius:
Re-tagging is no-go

gianarb:
I have submitted this PR because I don't know what is a good flow to resolve this problem.. :)

can I update this PR with 0.9.1 and tag release?
is it good about you? :)
Thanks

gianarb:
I don't know why but the tests are broken.. :) I can try to find the problem

PR #417, opened 2015-06-04
Bump some dependencies versions to be ZF 2.5 compliant

mpalourdio:
Because of phpunit set to ~3.7, DoctrineORMModule was stuck to ZF 2.4.*
PHPCodeSniffer has been updated to ~2 too.


gianarb:
Thanks a lot!

PR #418, opened 2015-06-04
Change build paremeters to prevent errors with hhvm and php7

mpalourdio:
See https://github.com/doctrine/DoctrineModule/pull/512 for reference.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,12 @@ php:
   - 5.4
   - 5.5
   - 5.6
-  - 7.0
+  - 7
   - hhvm
 
 matrix:
   allow_failures:
-    - php: 7.0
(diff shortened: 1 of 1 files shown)


mpalourdio:
This PR could be useless if a workaround is found for ZF2. I've opened an issue here

mpalourdio:
Fixed in https://github.com/zendframework/zf2/pull/7571

mpalourdio:
I ended up with just leaving the commit that fixes the builds. It is stille relevant as with phpunit 4.*, hhvm reports a wrong convering percentage, and with php7, xdebug is not loaded, so no coverage at all.

gianarb:
Thanks

PR #428, opened 2015-08-30
metadatagrapher - fix diagram inconcistency

Nono1971:
Hi, 

There was a little inconcistency in the yUML class diagram : when an entity had several entities refering to it with ManyToOne and with the same reference name => if these relations were both bidirectional and moodirectional, the link was not correct, refering on the monodirectional side, the relation of the other entity... 

more easy for me to explain with images : 
bug

bug fixed

### DIFF
--- a/src/DoctrineORMModule/Yuml/MetadataGrapher.php
+++ b/src/DoctrineORMModule/Yuml/MetadataGrapher.php
@@ -114,8 +114,9 @@ class MetadataGrapher
             }
         } else {
             foreach ($class2->getAssociationNames() as $class2Side) {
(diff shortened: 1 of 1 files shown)


gianarb:
Thanks!

Ocramius:
This sort of change requires tests.

Nono1971:
No problem, I work on the test case and I push when ready :)

gianarb:
@Ocramius do I revert it?

Ocramius:
This is fine as-is: no revert needed, as @Nono1971 is writing the tests separately

PR #431, opened 2015-08-31
Fix version test to new Doctrine version

eminetto:


Ocramius:
@eminetto what was the error you were getting prior to this change?

eminetto:
I receive an error regarding to Second Level Cache, because it is not enabled.

eminetto:
To be more specific, 
```
$entityManager->getCache()
```
returns null

Ocramius:
@gianarb poke

gianarb:
Version::compare description is 
> Returns -1 if older, 0 if it is the same, 1 if version passed as argument is newer.

`Version::compare('2.5.0') <= 0` the correct check shoud be it..
What do you think @Ocramius @eminetto?

I'm working to enable this matrix #432  to test this PR

Ocramius:
#432 seems to be a better fix (with tests).

See comments there, closing here.

PR #432, opened 2015-08-31
Env doctrine/orm version

gianarb:


.travis.yml
```diff
@@ -7,6 +7,10 @@ php:
   - 7
   - hhvm
 
+env:
+  - DOCTRINE_ORM="~2.4.0"
+  - DOCTRINE_ORM="~2.5.0"
```
Ocramius: I'd rather bump the dependency to the ORM to 2.5 than introducing this sort of maintenance complexity.

gianarb: do you increase dependency to doctrine/orm: ~2.5 and stop?

Ocramius: Yes

gianarb:
Fixed by #434

PR #440, opened 2015-11-17
[WIP] Moving away from array_map due to error supression

dragoonis:
Due to this bug/feature in php: https://bugs.php.net/bug.php?id=55416

The error message you get is:
`An error occurred while invoking the map callback in Module.php on line 137`

Which is really just not useful or helpful. By switching to the foreach() then normal exception handling and backtrace output can resume.

### DIFF
--- a/src/DoctrineORMModule/Module.php
+++ b/src/DoctrineORMModule/Module.php
@@ -133,8 +133,10 @@ class Module implements
                 )
             );
         }
-
-        $cli->addCommands(array_map(array($serviceLocator, 'get'), $commands));
+        
+        foreach($commands as $command) {
+            $cli->add($serviceLocator->get($command));
+        }
 
         /* @var $entityManager \Doctrine\ORM\EntityManager */
         $entityManager = $serviceLocator->get('doctrine.entitymanager.orm_default');


gianarb:
Thanks

Ocramius:
This is a bootstrap-only concern, and the warnings are fixed in 7.* ;-)

dragoonis:
@Ocramius all current versions are running on 5.x, and thus this is a problem.

Are you happy to replace an array_map with a foreach() to help everyone on 5.x? (which is 100% of DoctrineORModule users atm)

Ocramius:
@dragoonis not really, tbh.

Basically, it seems like you should go through every PHP codebase with anything that may trigger any exception anywhere, and get rid of `array_map` there.

As mentioned already in this thread:
 - this is a bootstrap-only concern (not runtime), therefore you should fix whatever is triggering the previous exception, IMO
 - this bug is already solved in PHP 7.x

dragoonis:
@Ocramius 

`Basically, it seems like you should go through every PHP codebase with anything that may trigger any exception anywhere, and get rid of array_map there.`
There isn't any, it's only in DoctrineORMModule.

`this is a bootstrap-only concern`
it's still a problem.

` therefore you should fix whatever is triggering the previous exception, IMO`
without converting from array_map() to foreach() you don't know what the exception is, it gets supressed to "error when invoking array_map" - so you can't.

`this bug is already solved in PHP 7.x`
Yes, but what about this being solved for the the 100% production user base, who are on 5.5?

Ocramius:
> There isn't any array_map, it wasn't any of my code, it was doctrine bootstrapping bits, it's only in DoctrineORMModule.

I am talking about [these projects](https://github.com/search?l=php&q=code%3Aarray_map&type=Code&utf8=%E2%9C%93) here.
I am aware that `array_map` is broken in 5.x (learned about it too late as well), but it's also indeed fixed in the 7.x series, which means that this minor annoyance would be fixed there.

 > Yes, but what about this being solved for the the 100% production user base, who are on 5.5?

I'm fine with having a fix (for now), even though it really isn't a bug in this codebase (that's why I'm arguing in first place).
I would need a test case that triggers the exception though (and verifies that the caught exception matches the original one, and isn't a warning).

dragoonis:
`I'm fine with having a fix (for now), even though it really isn't a bug in this codebase (that's why I'm arguing in first place).`

It's not a bug in your code, but in PHP 5.x, so we need to adapt accordingly.

`I would need a test case that triggers the exception though (and verifies that the caught exception matches the original one, and isn't a warning).`

My use case was that I didn't put in any DB credentials in my module's config file, so a PDO connection exception got thrown.

If you put in invalid DB credentials in module config, and then run migrations command, it will be reproducible.

Ocramius:
@dragoonis yes, I am asking for an actual phpunit test case tho

dragoonis:
@Ocramius realistically I won't find the time to create a reproducible test case right now due to project commitments/deadlines and PPI / PHP-FIG commitments.

gianarb:
Really old PR, I close it. 

@dragoonis if you are interested to continue please ping me or open a new one! 
Thanks a lot

PR #443, opened 2015-12-07
Added ability to configure the version column in MigrationsConfigurationFactory

asgrim:
Depends on doctrine/migrations#383 being merged, this allows configuration of the migration column name via the MigrationsConfigurationFactory.

### DIFF
--- a/config/module.config.php
+++ b/config/module.config.php
@@ -169,6 +169,7 @@ return array(
                 'name'      => 'Doctrine Database Migrations',
                 'namespace' => 'DoctrineORMModule\Migrations',
                 'table'     => 'migrations',
+                'column'    => 'version',
             ),
         ),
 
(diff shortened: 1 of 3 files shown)


asgrim:
Now that doctrine/migrations#383 is merged, this should be ready. I've written the change in such a way that it shouldn't break if the method is not available `setMigrationsColumnName`.

gianarb:
Perfect! :+1: 

Feedback buy other users?? :)

asgrim:
BTW, doctrine/migrations has been tagged `1.2.0` including the merged PR doctrine/migrations#383 so this should be pretty much ready :)

gianarb:
No feedback :D We can merge!

gianarb:
Thanks a lot!

asgrim:
thank YOU! :)

PR #447, opened 2016-01-31
Up ZF to ^2.5

TomHAnderson:
In support of moving doctrine-module to zend-hydrator, this module cannot pin versions below 2.5

https://github.com/doctrine/DoctrineModule/pull/548

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 5.4
   - 5.5
   - 5.6
   - 7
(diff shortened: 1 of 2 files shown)


TomHAnderson:
Before this PR can move forward a decision must be made for which version of Zend Framework to support. https://github.com/doctrine/DoctrineMongoODMModule/blob/master/composer.json uses "2.*" which I think is a better tag.

stijnhau:
I think this can be closed because it support the v3 now.

PR #457, opened 2016-03-21
tests fails on all php versions except 5.4 becouse of change in zend-mvc 2.7 version

svycka:


svycka:
Looks like BC issue with ServiceManager no?

ping @weierophinney @Ocramius

weierophinney:
Looks like the tests are attempting to override existing services; you likely need to add a call to setAllowOverride during your setup.

svycka:
@weierophinney found when this changed not sure if this is BC break but it happened in zend-mvc 2.7.0 it was added here https://github.com/zendframework/zend-mvc/blob/release-2.7.0/src/Service/ServiceManagerConfig.php#L52

gianarb:
Yes, I fixed it for DoctrineModule.. I proceed to do it for all when I have a chance..

svycka:
@gianarb I think this module has more issues because with `composer update --prefer-lowest` tests also fail but they fail with and without this fix.

gianarb:
@svycka thanks!

PR #459, opened 2016-03-22
Added ServiceManager v3 support

svycka:


Slamdunk:
Hi, for sure this is not the right place, but I'm trowing this here because it's easy to see the flaw.

The dependancy of DoctrineORMModule to ZendDeveloperTools it's a pain, isn't it?

DoctrineModule and DoctrineORMModule could be easily upgraded following their *production* dependancies, but we are stuck by a development dependancy that is totally optional.

Moreover, ZendDeveloperTools has a dependancy on modules that are neither in DoctrineModule nor in DoctrineORMModule (zend-eventmanager), so the final result is that 2 widespread modules with few dependancies are stopped from evolving by an optional package full of non-related dependancies.

We should remove all code related to ZendDeveloperTools from here and move it to another bridge module: in this middle-module we can stop version upgrading, but if the deveolper doesn't use it, he (and we) could follow a faster upgrade.

What do you think?

svycka:
I agree but this would be BC break unless we could remove only tests :D

gianarb:
I am working on a good plane for the future.. Can you help me with your point of view please?
https://github.com/doctrine/doctrine-website-sphinx/pull/125

About the split yes, it's a BC break we can not do it now @Slamdunk but I think that it's good for DoctrineORMModule 2.0, in first because I am not sure about the maintenance plane for **ZendDeveloperTools** and for this reason I like the idea to push out this dependencies..

svycka:
@gianarb yes I can can change those I used it just for testing purposes because nor ZendDeveloperTools not DoctrineModule does not support SMv3 at the moment

composer.json
```diff
@@ -31,20 +31,38 @@
             "email": "guilhermeblanco@hotmail.com"
         }
     ],
+    "repositories": [
```
Ocramius: To be removed pre-merge

composer.json
```diff
@@ -31,20 +31,38 @@
...
+        },
+        {
+            "type": "vcs",
+            "url": "https://github.com/svycka/ZendDeveloperTools"
+        }
+    ],
     "require":           {
         "php":                               "^5.5 || ^7.0",
-        "doctrine/doctrine-module":          "~1.0",
+        "doctrine/doctrine-module":          "dev-feature/sm3-support",
```
Ocramius: To be removed pre-merge

composer.json
```diff
@@ -31,20 +31,38 @@
...
+        "zendframework/zend-config":          "^2.6",
+        "zendframework/zend-view":            "^2.5.2",
+        "zendframework/zend-serializer":      "^2.6",
+        "zendframework/zend-log":             "^2.5.2",
+        "zendframework/zend-i18n":            "^2.6",
+        "zendframework/zend-console":         "^2.6",
+        "zendframework/zend-code":            "^2.5.2",
         "doctrine/data-fixtures":             "1.0.*",
-        "zendframework/zend-developer-tools": "*",
+        "zendframework/zend-developer-tools": "dev-feature/sm-v3-support",
```
Ocramius: To be removed pre-merge

src/DoctrineORMModule/Form/Annotation/AnnotationBuilder.php
```diff
@@ -56,7 +56,8 @@ public function setEventManager(EventManagerInterface $events)
     {
         parent::setEventManager($events);
 
-        $this->getEventManager()->attach(new ElementAnnotationsListener($this->objectManager));
+        $annotationsListener = new ElementAnnotationsListener($this->objectManager);
+        $annotationsListener->attach($this->getEventManager());
```
Ocramius: Can be inlined: `(new ElementAnnotationsListener($this->objectManager))->attach($this->getEventManager());`

src/DoctrineORMModule/Form/Annotation/ElementAnnotationsListener.php
```diff
@@ -48,7 +48,7 @@ public function __construct(ObjectManager $objectManager)
     /**
      * {@inheritDoc}
      */
-    public function attach(EventManagerInterface $events)
+    public function attach(EventManagerInterface $events, $priority = 1)
```
Ocramius: is `$priority` added by the interface?

src/DoctrineORMModule/Service/DoctrineObjectHydratorFactory.php
```diff
@@ -27,9 +28,16 @@ class DoctrineObjectHydratorFactory implements FactoryInterface
     /**
      * {@inheritDoc}
      */
-    public function createService(ServiceLocatorInterface $serviceLocator)
+    public function __invoke(ContainerInterface $container, $requestedName, array $options = null)
     {
-        $parentLocator = $serviceLocator->getServiceLocator();
```
Ocramius: Wouldn't this break 2.x compat?

tests/DoctrineORMModuleTest/Yuml/YumlControllerTest.php
```diff
@@ -54,7 +54,7 @@ public function setUp()
     {
         $this->httpClient     = $this->getMock('Zend\\Http\\Client');
         $this->controller     = new YumlController($this->httpClient);
-        $this->pluginManager  = $this->getMock('Zend\\Mvc\\Controller\\PluginManager');
+        $this->pluginManager  = $this->getMock('Zend\\Mvc\\Controller\\PluginManager', array(), array(), '', false);
```
Ocramius: short arrays

tests/DoctrineORMModuleTest/Service/FormAnnotationBuilderFactoryTest.php
```diff
@@ -36,7 +36,7 @@ class FormAnnotationBuilderFactoryTest extends PHPUnit_Framework_TestCase
     public function testFormElementManagerGetsInjected()
     {
         $entityManager      = $this->getMock('Doctrine\ORM\EntityManager', array(), array(), '', false);
-        $formElementManager = $this->getMock('Zend\Form\FormElementManager');
+        $formElementManager = $this->getMock('Zend\Form\FormElementManager', array(), array(), '', false);
```
Ocramius: short arrays

Ocramius:
At first glance, I see a lot of code in this PR (and doctrine/DoctrineModule#558) that is just due to keeping compat with servicemanager v2.

Do we need that?

src/DoctrineORMModule/Form/Annotation/ElementAnnotationsListener.php
svycka: yes

src/DoctrineORMModule/Service/DoctrineObjectHydratorFactory.php
svycka: nop, I have moved `$serviceLocator->getServiceLocator()` into `createService()` method and SMv3 uses `__invoke()` always and $container allways will be ServiceManager instance not a plugin manager.

svycka:
Please at least one release with SMv3 and SMv2 then we can remove SMv2 support. Adding zend-hydrator anyway will require break BC so then you could remove SMv2 support.

svycka:
I can remove my forks now or someone will do this pre-merge?

Ocramius:
@svycka if svycka/DoctrineORMModule#04db2d2 passes, you can add another commit removing the workarounds, and we can then merge

samsonasik:
@Ocramius @svycka removed the workarounds

webimpress:
It would be nice to update this PR to support also [zend-mvc](https://github.com/zendframework/zend-mvc) and [zend-stdlib](https://github.com/zendframework/zend-stdlib) v3.

svycka:
@webimpress this is not possible because of zend-hydrator requirement. Would be BC break. but I am ok with this BC break :D if could add zf3 support

webimpress:
@svycka finally one day this library needs support ZF3 modules, as it is now it's not possible to migrate application to zf3 modules. It is not my decision when it will be done, and what version will BC break.

@Ocramius @samsonasik can you tell us what is the plan of releases of this library? when it will support zf3 modules?

MidnightDesign:
I don't get it. What's the status? Is there still SM2 code left to remove? Do we want to upgrade mvc and stdlib too? Is there something I can do to help?

igormx:
The problem is only with SMv3 or with another zf3 components?

tmcnicholls:
Now that ZF3 has been released, are there any plans to add support for it?

https://framework.zend.com/blog/2016-06-28-zend-framework-3.html

gianarb:
Thanks @svycka

PR #430, opened 2015-08-31
Medatagrapher fix

Nono1971:
Hi Ocramius,

Well... after My last pull request, I created the test as you asked... and the testCase found another bug :) 
=> depending of metadata's order, there was another inconcistency : 
If the inversed side associations were visited first, we had a double arrow for the relation...

Fixed it too :)

I must tell you that I took inspiration from your YumlController to create a Symfony bundle... I Missed yUML diagram so much when I started using Doctrine with CodeIgniter (first) and now with Symfony....

https://packagist.org/packages/onurb/doctrine-yuml-bundle

I first reused in my bundle your MetadataGrapher class as is (only changed the namespace) : I didn't want to load DoctrineORMModule and all its dependencies for only one class !... (the only problem for me is now to write code twice to push you my fixes ;) )

I think it's fair to inform you that you're work has been reused :)

Thank's Marco for you're work on Doctrine (and all the rest of the team of course)... I just love this library !.

Bruno.

PR #476, opened 2016-07-03
externalize MetadataGrapher to a library

Nono1971:
Hi, as suggested By Ocramius few month ago... I propose to externalize the Yuml MetadataGrapher into an external library.

I left it behind (and almost forgot it)... but I have decided refactoring my class again... that instead of modifiing both sides again... It's infact really better to finaly do it :)
 You can go there to check if the library pleases to you : https://github.com/Nono1971/Doctrine-MetadataGrapher


gianarb:
@Ocramius pleae give some feedback about this idea.. It's a next external dependencies.. I don't really like that but if you like the idea we can merge :)

Thanks @Nono1971 for your effort!

Nono1971:
for your help, it was suggested here : #430

Nono1971:
Well... I just found a bug not covered by tests into the library... wich is not present without my refactoring... 

 I'll work on it after my job... don't merge now ;)

Nono1971:
Hello,

I worked on it... and found a solution... but I have a problem testing it... You might help maybe ^^ 

let's describe the problem.

The bug came from the Metadata::getClassReverseAssociationName() method... Wich was the same used here before my fix : 

    private function getClassReverseAssociationName(ClassMetadata $class1, ClassMetadata $class2)
    {
        foreach ($class2->getAssociationNames() as $class2Side) {
            $targetClass = $this->getClassByName($class2->getAssociationTargetClass($class2Side));
            if ($class1->getName() === $targetClass->getName()) {
                return $class2Side;
            }
        }

        return null;
    }

I replaced the failing method by this : 

    /**
     * @param \Doctrine\Common\Persistence\Mapping\ClassMetadata $class1
     * @param string $association
     * @return string|null
     */
    private function getClassReverseAssociationName(ClassMetadata $class1, $association)
    {
        if ($class1->getAssociationMapping($association)['isOwningSide']) {
            return $class1->getAssociationMapping($association)['inversedBy'];
        } else {
            return $class1->getAssociationMapping($association)['mappedBy'];
        }
        return null;
    }

Of course I changed also the call of the method to replace the class by the association name into getAssociationString() method : 

    $class2SideName = $this->getClassReverseAssociationName($class1, $association);


It works perfectly ! :)

Now are the troubles comming... Tests !!

I use the ClassMetadataInfo::getAssociationMapping() method, wich was not used before... and I'd like to test it correctly... but this method is not in the interface Doctrine\Common\Persistence\Mapping\ClassMetadata so I can't mock the method without changing the interface... wich is in doctrine/common... (and also implemented into doctrine/odm wich should be fixed if the interface would change : the doctrine/odm ClassMetadataInfo doesn't implements this method yet of course ^^).

I pass all tests changing manually the interface doctrine/common... only adding the getClassReverseAssociationName() method.

Do you think the interface could be changed on doctrine/common ? 
should I make a pull request there or would you as it fixes the bug here to ? 
=> I make a push with the failing testcase here... (I'll edit to link the pull request))

If the interface could be changed, I could version my package as it would fix bug here too :)

Nono1971:
I pushed the bug testCase (and partial fix) here #482

Nono1971:
I applied my travis ugly fix in my repo, so all tests are good... I version it and you can merge :)

Nono1971:
Damn, After ocramius remark here #482 , I look for the tree... and I realise that I really started from too far, forgotting to pull changes on master before creating the branch... I close this pull request and create a clean new one, with a package up to date

Ocramius:
Moved to #485

PR #482, opened 2016-07-12
Bug/metadata grapher test case

Nono1971:
As I explained in #476 I found a bug in MetadataGrapher... travis should be red :D 

I first push the testcase of the bug (as I explained, I have a part of the solution... hope you can help to complete it :)


Nono1971:
I fixes the bug, and adapted the tests with the new method used.

Tests are still down with travis
To mock the method in tests, the doctrine/common has to be modified... I did it manually, and tests are working fine then :)

The change is : 
    // add into interface Doctrine\Common\Persistence\Mapping\ClassMetadata 
       
    /**
     * Gets the mapping of an association.
     *
     * @param string $fieldName The field name that represents the association in
     *                          the object model.
     *
     * @return array The mapping.
     *
     * @throws MappingException
     */
    public function getAssociationMapping($fieldName);

I could make travis test pass with a uggly script to replace the file by a modified one... but it would be better if it was merged into doctrine/common.

as I said in #476 mongodb-odm also implements this interface in Doctrine\ODM\MongoDB\Mapping\ClassMetadataInfo and would also have to be fixed if this method would be added to the interface (and maybe the same in others packages I would not think or know about ?^^)

.travis.yml
```diff
@@ -12,6 +12,8 @@ matrix:
 
 before_script:
   - composer update --prefer-dist
+  - rm -f ./vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/ClassMetadata.php
```
Ocramius: To be removed

.travis.yml
```diff
@@ -12,6 +12,8 @@ matrix:
...
+  - mv ./build/ClassMetadata.php.travis.dist ./vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/ClassMetadata.php
```
Ocramius: To be removed

build/ClassMetadata.php.travis.dist
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * Contract for a Doctrine persistence layer ClassMetadata class to implement.
+ *
+ * @link   www.doctrine-project.org
+ * @since  2.1
+ * @author Benjamin Eberlei <kontakt@beberlei.de>
+ * @author Jonathan Wage <jonwage@gmail.com>
+ */
+interface ClassMetadata
```
Ocramius: This class is NOT to be provided by this module

tests/DoctrineORMModuleTest/Yuml/MetadataGrapherTest.php
```diff
@@ -143,11 +164,16 @@ public function testDrawOneToOneBiDirectionalInverseAssociation()
...
+            'isOwningSide' => true,
+            'mappedBy'      => null,
+            'inversedBy'    => 'a'
+        )));
         $class2->expects($this->any())->method('isAssociationInverseSide')->will($this->returnValue(false));
         $class2->expects($this->any())->method('isCollectionValuedAssociation')->will($this->returnValue(false));
         $class2->expects($this->any())->method('getFieldNames')->will($this->returnValue(array()));
-
-        $this->assertSame('[A]<a 1-b 1<>[B]', $this->grapher->generateFromMetadata(array($class1, $class2)));
+        $expected = "[A]<a 1-b 1<>[B]";
```
Ocramius: Revert

Ocramius:
The patch as-is is a no-go. Can you please only modify the tests to expose the bug first? If the test suite crashes, that is not a problem, but embedding the interface is absolutely a no-go, as it would hide future BC breaks.

Nono1971:
Hi Ocramius,

I agree with you, travis patch, replacing the ClassMetadata Interface is to avoid... this is temporary the only way I found to make test pass... and I'm not good with it. I think interface should have to be corrected in doctrine/common instead of patching it like this here... but it could break other packages implementing it... not simple ^^

Let me explain the bug testcase : 
two classes, having two bidirectional relations with dirrerents names. => the mapping creates a third link between classes.

For this testcase
Class CDis Owner of both relation
Class AB::c refers Class CD,  mappedBy= "a"
Class AB::d refers Class CD, mappedBy= "b"

Class CD::a refers to Class AB, inversedBy="c"
Class CD::b refers to Class AB, inversedBy="d

instead of drawing two lines like this : [AB]<a 1-c *<>[CD],[AB]<b 1-d *<>[CD]
it creates a third one linking CD::b to AB::c => Relation that doesn't exist.
 : [AB]<a 1-c *<>[CD],[AB]<a 1-d *<>[CD],[CD]<>c *-b 1>[AB]

I pushed by steps to can check the bug testcase. and to illustrate the bug I found... so you can check... and I hope help, as I'm blocked with tests.

to get the failing test only :
    git checkout e4bc990 

You can read #476, I explained my testing problem there too.

By the way, I pushed here to give an example, more than merging it. the goal is to find a way to do it... I know it is two ugly to be merged... (As shown in my last Commit description)

If you know a way to mock the method in tests without replacing the interface, I'd prefer too ;) (I explained my trouble in #476 wich is the pull request I would like to merge instead of this one... WHEN TESTS WON'T BE NO MORE SO UGLY on the external repo...

Ocramius:
> interface should have to be corrected in doctrine/common

Not possible: interface changes are massive BC breaks

.gitattributes
```diff
@@ -0,0 +1,7 @@
+/.travis export-ignore
```
Ocramius: Change is unrelated? Did you rebase your branch onto latest `master`?

src/DoctrineORMModule/Yuml/MetadataGrapher.php
```diff
@@ -142,19 +142,16 @@ private function getAssociationString(ClassMetadata $class1, $association)
...
+    private function getClassReverseAssociationName(ClassMetadata $class1, $association)
     {
-        foreach ($class2->getAssociationNames() as $class2Side) {
-            $targetClass = $this->getClassByName($class2->getAssociationTargetClass($class2Side));
-            if ($class1->getName() === $targetClass->getName()) {
-                return $class2Side;
-            }
+        if ($class1->getAssociationMapping($association)['isOwningSide']) {
+            return $class1->getAssociationMapping($association)['inversedBy'];
+        } else {
```
Ocramius: No need for the `else` clause here

Nono1971:
I managed to find a solution... reading that F** manual in train from work ;)

I finaly used the mockBuilder to add the method to the mock object... Great ! (I'm still a unit tests beginner ^^)

Well, infact, I propose you, as you suggested 10 month ago, when I fixed a first MetadataGrapher bug #430 to externalize it in a separate library, so we could use it as well for DoctrineORMModule and for the symfony bundle I made.

I pushed the changes to pass tests... but It should not be merged if you agree with the external solution => I made another pull request #476  to make this externalization (here I just fixed the bugged method, but I completely refactored the MetadataGrapher on my repo to avoid multiples foreach, storing metadata and ClassStrings in arrays to generate it only once... 

Sorry for the upstream/master merge that I forgot ^^ (I'll check for #476 to rectify too ;) )
and thanks for the useless else I missed (removed) :)

Nono1971:
I think we can close this PR, 
First, I opened #485 to externalize the solution 
Second, this patch was made from an outdated master... you'll never merge that ;) )

PR #483, opened 2016-07-14
Externalize/metadata grapher

Nono1971:
Following #476 I made my changes from an up to date version...

This is about externalizing the MetadataGrapher, so could be used as well by My symfony's onurb/doctrine-yuml-bundle and DoctrineOrmModule :)

@Ocramius suggested that here #430 I finaly did it, after refactoring the class... so, if you always agree with that, let's go ! :)
You can go check the code here : https://github.com/Nono1971/Doctrine-MetadataGrapher.
I implemented the same tests that where used here (modified after the new bug I found next week and the patch used here : #482) .

PR #485, opened 2016-07-14
externalize MetadataGrapher

Nono1971:
I hope code is really up to date this time :D (after #476 and #483 fails)


view/zend-developer-tools/toolbar/doctrine-orm-mappings.phtml
```diff
@@ -19,7 +19,7 @@
         </span>
         <form action="<?php echo $this->url('doctrine_orm_module_yuml'); ?>" method="post" target="_blank">
             <?php
-                $grapher = new \DoctrineORMModule\Yuml\MetadataGrapher();
+                $grapher = new \Doctrine\ORM\MetadataGrapher\YUMLMetadataGrapher();
```
Ocramius: Why is the metadata grapher in the `Doctrine\ORM` namespace?

composer.json
```diff
@@ -40,7 +40,8 @@
         "zendframework/zend-hydrator":       "^1.1",
         "zendframework/zend-mvc":            "^2.5.2",
         "zendframework/zend-servicemanager": "^2.7.5 || ^3.0.3",
-        "symfony/console":                   "~2.5|~3.0"
+        "symfony/console":                   "~2.5|~3.0",
+        "onurb/doctrine-metadata-grapher":   "~1.0"
```
Ocramius: I like the idea of getting rid of one piece of maintenance here, and to stash it away, but will you keep up with the maintenance?

I'm not questioning your technical skills, I'm just asking how cautious you will be with upgrades and changes (semver-wise).

Nono1971:
I think maintenance shouldn't be a problem.

I'm not a fool, If I want to make big changes, I can test it localy on its own branch, and talk about it with you (and make you test) before applying any change... MetadataGrapher is your idea first... I'm not here to break anything.
I can also give you full access on this repo :) (just to find how first :D)

The only thing that I think could be added to the MetadataGrapher in the future, would be to show columns types in mapping... but it's already great like that, no emergency ! The MetadataGrapher should stay stable a while :)

Nono1971:
for the Doctrine\ORM namespace, I thought it was appropriate, because works only with the ORM... and I was not inspired. 
If you have other suggestion, it's not to late to change namespace ;)

Ocramius:
> I can also give you full access on this repo

We're trying to reduce development burden, not to increase it :P

 > I'm not a fool, If I want to make big changes, I can test it localy on its own branch, and talk about it with you (and make you test) before applying any change... MetadataGrapher is your idea first... I'm not here to break anything.

Make sure that everything will be kept [semver](http://semver.org/) compliant and it should be fine.

 > or the Doctrine\ORM namespace, I thought it was appropriate, because works only with the ORM... and I was not inspired. 

You should use your own namespace. `Nono1971\MetadataGrapher` is a good example of this.

Nono1971:
OK, I look for semver one more thing to learn about :)
for the namespace, I'll use Onurb\ as my vendor name is already for my bundle. I don't like numbers in namespaces ;) (didn't even tried if it is allowed by php :D)

I change the namespace and fix my symfony bundle dependency... apply changes here to and let's go ^^.

Nono1971:
namespace changed :)

Nono1971:
Hi, @Ocramius,

I have big news on my side :)

I first completely refactored the MetadataGrapher (I reached 9.96 on scrutinizer :) ).  By the way, I found a new bug in inheritance (fields of the parent, of the parent were ignored... It is fixed in my next release, I now check all the inheritance tree if exists).

Infact, I  did a little more than that last weeks... I managed to add optional functionalities to use a little bit more of the yuml API.... almost everything infact ;) and to allow final user usage with annotations or configuration (to define in my Symfony bundle or in this Zend Module (I'll do it and update this PR quick).

Here are the new options : 
=> color Entities in the mapping (or set a namespace default color)
=> add notes on Entities
=> show / hide fields properties
=> show selected methods of Entities (useless for getters / setters, but great to present logical methods in the map)

This is fully optional : you can call the generateMetadata() method as before and user can do some stuff with annotations on entities or entities methods... and there are only few adjustments to do here to use namespaces default configuration (like default hide/show the entire mapping, or define a color for a complete module or any sub namespace).

The code is still 100% covered (I really enjoy TDD now :)) ... and becomes really powerfull to allow users to display the map as fine tuned as possible.

It's almost ready, the only think I still have to decide, is to define how to display attributes with properties !!! :D not the way to do it, it works perfectly... but the way it should look to be nice :) => not a big deal ;) (I work on it after my job, in train, at night or WE... everything should be ready in only a day or two... 

Infact, cause it will be used here as well and I can't decide myself... I come here to take advice ;)


I' like to display the type of the field, the length (if string)... and maybe also to add scale and precision to decimal types.. (if you have other suggestions of particular types with interesting properties related I would have forgotten... :)) ... and I'd also like to have a tag for unique fields.... 

It would be great to decide together the final shape so it pleases everyone :)

Here is a yuml done with examples of what I thought about... please tell me your favorite... or if you see another string for the fields that what those I imagined. http://yuml.me/23da48bf


Don't merge now, I'll have to make some adaptations to the module_config and to the view so final user can configure it on his will.

I think the code will be final this time, so the library works great, code is clean, logic well separated... and yuml usage is almost complete :) what do next ?

Thanks for your attention :) I wait for your opinion for the field style (not just ocramius , the most opinions we get, the best it is...)

gianarb:
@Ocramius news about this?
I am not really sure that have a new dependency can help.. but it depends of the complexity that we are in our mind for the future.

TomHAnderson:
I'll close this after a year from it's creation if no work is done from here.

PR #479, opened 2016-07-04
ZF3 support

webimpress:
Full ZF3 support. Dropped support PHP 5.5 and ZF2.
It works with https://github.com/doctrine/DoctrineModule/pull/564.

Need a bit more work.
Documentation needs update.

Any suggestions would be appreciated.

(library uses now PSR-4 autoloading, as mentioned in https://github.com/doctrine/DoctrineModule/pull/564 doctrine uses deprecated PSR-0, it could be moved back to PSR-0 later on if needed)

PR #488, opened 2016-07-17
Fix: Streamline configuration examples

localheinz:
This PR

* [x] streamlines configuration examples in the documentation

Somewhat related to #479.

💁 Some examples return arrays, some examples show slices of arrays - it's probably better to be consistent across the documentation.


docs/migrations.md
```diff
@@ -13,5 +13,5 @@ return array(
             ),
         ),
     ),
-)
+);
```
localheinz: This example is missing a trailing `;` to turn this into a valid statement.

gianarb:
Thanks

PR #492, opened 2016-07-20
Update to EXTRA_ORM.md to correct config key

DannyvdSluijs:
As there is no reference to be found to my_dbal_default I found that the correct value needs to be orm_default, as goes for all the other configuration elements.


gianarb:
Thanks @DannyvdSluijs

PR #494, opened 2016-08-01
Hotfix router config

thomasvargiu:
- Fixed route to be compatible with both `zend-mvc`<3.0 and `zend-router`. The route configuration is moved to the Module class implementing `RouteProviderInterface`.

- Replaced `Config` key with the standardized and lowercased `config` key.

### DIFF
--- a/config/controllers.config.php
+++ b/config/controllers.config.php
@@ -17,29 +17,15 @@
  * <http://www.doctrine-project.org>.
  */
 
-use DoctrineORMModule\Yuml\YumlController;
-use Zend\Http\Client;
-use Zend\ServiceManager\AbstractPluginManager;
-use Zend\ServiceManager\Exception\ServiceNotFoundException;
+namespace DoctrineORMModule;
 
-return array(
-    'factories' => array(
+use DoctrineORMModule\Yuml\YumlController;
+use DoctrineORMModule\Yuml\YumlControllerFactory;
+
+return [
+    'factories' => [
         // Yuml controller, used to generate Yuml graphs since
         // yuml.me doesn't do redirects on its own
-        'DoctrineORMModule\\Yuml\\YumlController'  => function (AbstractPluginManager $pluginManager) {
-            $config = $pluginManager->getServiceLocator()->get('Config');
-
-            if (! isset($config['zenddevelopertools']['toolbar']['enabled'])
-                || !$config['zenddevelopertools']['toolbar']['enabled']
-            ) {
-                throw new ServiceNotFoundException(
-                    'Service DoctrineORMModule\\Yuml\\YumlController could not be found'
-                );
-            }
-
-            return new YumlController(
-                new Client('http://yuml.me/diagram/plain/class/', array('timeout' => 30))
-            );
-        },
-    ),
-);
+        YumlController::class => YumlControllerFactory::class,
+    ],
+];
--- a/config/module.config.php
+++ b/config/module.config.php
@@ -257,7 +257,7 @@ return array(
     'router' => array(
         'routes' => array(
             'doctrine_orm_module_yuml' => array(
-                'type' => 'Zend\\Mvc\\Router\\Http\\Literal',
+                'type' => 'literal',
                 'options' => array(
                     'route' => '/ocra_service_manager_yuml',
                     'defaults' => array(
--- a/src/DoctrineORMModule/Yuml/YumlControllerFactory.php
+++ b/src/DoctrineORMModule/Yuml/YumlControllerFactory.php
@@ -1,0 +1,77 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace DoctrineORMModule\Yuml;
+
+use Interop\Container\ContainerInterface;
+use Interop\Container\Exception\ContainerException;
+use Zend\Http\Client;
+use Zend\ServiceManager\AbstractPluginManager;
+use Zend\ServiceManager\Exception\ServiceNotFoundException;
+use Zend\ServiceManager\FactoryInterface;
+use Zend\ServiceManager\ServiceLocatorInterface;
+
+class YumlControllerFactory implements FactoryInterface
+{
+    /**
+     * Create service
+     *
+     * @param ServiceLocatorInterface $serviceLocator
+     *
+     * @return YumlController
+     */
+    public function createService(ServiceLocatorInterface $serviceLocator)
+    {
+        if ($serviceLocator instanceof AbstractPluginManager) {
+            $serviceLocator = $serviceLocator->getServiceLocator() ?: $serviceLocator;
+        }
+
+        return $this($serviceLocator, YumlController::class);
+    }
+
+    /**
+     * Create an object
+     *
+     * @param  ContainerInterface $container
+     * @param  string             $requestedName
+     * @param  null|array         $options
+     *
+     * @return YumlController
+     *
+     * @throws \Interop\Container\Exception\NotFoundException
+     * @throws ServiceNotFoundException if unable to resolve the service
+     * @throws ContainerException if any other error occurs
+     */
+    public function __invoke(ContainerInterface $container, $requestedName, array $options = null)
+    {
+        $config = $container->get('config');
+
+        if (!isset($config['zenddevelopertools']['toolbar']['enabled'])
+            || !$config['zenddevelopertools']['toolbar']['enabled']
+        ) {
+            throw new ServiceNotFoundException(
+                sprintf('Service %s could not be found', YumlController::class)
+            );
+        }
+
+        return new YumlController(
+            new Client('http://yuml.me/diagram/plain/class/', ['timeout' => 30])
+        );
+    }
+}
--- a/tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
+++ b/tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
@@ -1,0 +1,97 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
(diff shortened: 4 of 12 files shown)


config/controllers.config.php
```diff
@@ -17,29 +17,15 @@
...
-                throw new ServiceNotFoundException(
-                    'Service DoctrineORMModule\\Yuml\\YumlController could not be found'
-                );
-            }
-
-            return new YumlController(
-                new Client('http://yuml.me/diagram/plain/class/', array('timeout' => 30))
-            );
-        },
+        Yuml\YumlController::class  => Yuml\YumlControllerFactory::class
```
webimpress: one space before `=>` please

webimpress: and trailing comma will be nice

config/module.config.php
```diff
@@ -254,21 +254,6 @@
     // zend developer tools                                           //
     ////////////////////////////////////////////////////////////////////
 
-    'router' => array(
-        'routes' => array(
-            'doctrine_orm_module_yuml' => array(
-                'type' => 'Zend\\Mvc\\Router\\Http\\Literal',
```
webimpress: we could keep it here, and use just `literal` here

src/DoctrineORMModule/Module.php
```diff
@@ -149,4 +151,33 @@ public function initializeConsole(EventInterface $event)
...
         $helperSet->set(new EntityManagerHelper($entityManager), 'em');
     }
+
+    /**
+     * Expected to return \Zend\ServiceManager\Config object or array to
+     * seed such an object.
+     *
+     * @return array|\Zend\ServiceManager\Config
+     */
+    public function getRouteConfig()
```
webimpress: as above, we should keep it in `module.config.php` (this way it is always done in zend/apigility modules)

tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
```diff
@@ -0,0 +1,39 @@
...
+        $serviceLocator->expects(static::any())->method('get')->with('config')->willReturn($config);
+
+        $pluginManager->expects(static::once())
+            ->method('getServiceLocator')
+            ->willReturn($serviceLocator);
+
+        $factory = new YumlControllerFactory();
+        $controller = $factory->createService($pluginManager);
+
+        static::assertInstanceOf(YumlController::class, $controller);
```
webimpress: I know it is a problem with the latest phpstorm release, but usually we are using `$this->` in phpunit. And probably it this whole library it is with `$this->` so maybe it should be changed, what do you think?

webimpress:
@thomasvargiu thanks for your fixes, I saw your comment on my PR but I didn't have chance to fix it before merge, thanks again!

tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
```diff
@@ -0,0 +1,39 @@
...
+            ]
+        ];
+
+        $serviceLocator = $this->getMockBuilder(ServiceLocatorInterface::class)
+            ->getMock();
+        $pluginManager = $this->getMockBuilder(AbstractPluginManager::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $serviceLocator->expects(static::any())->method('get')->with('config')->willReturn($config);
```
webimpress: maybe we can use `prophesize` ?

src/DoctrineORMModule/Yuml/YumlControllerFactory.php
```diff
@@ -0,0 +1,60 @@
...
+use Zend\ServiceManager\AbstractPluginManager;
+use Zend\ServiceManager\Exception\ServiceNotCreatedException;
+use Zend\ServiceManager\Exception\ServiceNotFoundException;
+use Zend\ServiceManager\FactoryInterface;
+use Zend\ServiceManager\ServiceLocatorInterface;
+use Zend\Http\Client;
+
+class YumlControllerFactory implements FactoryInterface
+{
+
```
webimpress: please remove above empty line

src/DoctrineORMModule/Yuml/YumlControllerFactory.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
```
webimpress: please add licence text as it is in all files

tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
```
webimpress: licence text please

tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
thomasvargiu: @webimpress It's not a problem, `assertInstanceOf` is a static method

webimpress: I know, but this is just global convention, will be nice to be consistent in the whole library.

https://blog.jetbrains.com/phpstorm/2016/07/completion-changes-in-phpstorm/

tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
thomasvargiu: I'd prefer `prophesize` but in the whole project is used the mock builder.

webimpress:
@thomasvargiu tests are failing because of wrong locked file, my locked file was generated for ZF2 modules, your is probably for ZF3 modules. Please check travis configuration (to pass tests on ZF3 `zendframework/zend-mvc-console` must be installed, and it's conflicted with ZF2, it is added to `suggest` section).

config/controllers.config.php
```diff
@@ -17,29 +17,15 @@
...
-                );
-            }
-
-            return new YumlController(
-                new Client('http://yuml.me/diagram/plain/class/', array('timeout' => 30))
-            );
-        },
-    ),
-);
+        YumlController::class => YumlControllerFactory::class
```
webimpress: trailing comma please

tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
webimpress: is then `->expects(static::any())` required there? (in this case many "invoked zero or more times" so completely doesn't matter)

thomasvargiu:
@webimpress Can I change the `doctrine-module` requirement do `dev-master`? Was your branch merged?

webimpress:
> Can I change the doctrine-module requirement do dev-master? Was your branch merged?

yeah, it's merged https://github.com/doctrine/DoctrineModule/pull/567
I'm going to check it soon and maybe prepare some fixes if needed.

tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
webimpress: @thomasvargiu so why you are consistent in one place (with mocking) and inconsistent in another (`static::`) with the whole library ?

thomasvargiu: @webimpress  
TestCase methods are instance methods:

https://github.com/sebastianbergmann/phpunit/blob/5.4.8/src/Framework/TestCase.php#L1509

Assert methods are static methods:

https://github.com/sebastianbergmann/phpunit/blob/5.4.8/src/Framework/Assert.php#L1320

Am I wrong using ```static::``` or it's a common mistake to use ```$this->``` for static methods? :)

tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
thomasvargiu: IMHO it's not important for the test scope, it would be just a constraint.  
But probably I should integrate more tests on this factory. I'm going to write those tests.

tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
webimpress: You are completely right, I know the method is `static` and could be used with `static::` but author of the PHPUnit suggest to use it `$this->` and you can check all examples in PHPUnit docs.
Also in this library all other tests are with `$this->` and greater majority of all other php libraries on github! So maybe everyone is wrong, but you can notice how many people who are using PHPStorm notice it, and finally it's gonna be changed in PHPStorm (please read blog post I put above).

I'd like just have CONSISTENT code in the library. :)

thomasvargiu: @webimpress Nice discussion! I've converted using `$this->`. You're right, it's better to have consistent code.
But I still have doubts on using `$this->` on static methods :)

webimpress: @thomasvargiu thanks. I don't know what is the original reason of using `$this->` not static by @sebastianbergmann - it is like that everywhere !!
Please have a look on PHPUnit tests in the phpunit repository: https://github.com/sebastianbergmann/phpunit/tree/master/tests

tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
```diff
@@ -0,0 +1,39 @@
...
+        $config = [
+            'zenddevelopertools' => [
+                'toolbar' => [
+                    'enabled' => true
+                ]
+            ]
+        ];
+
+        $serviceLocator = $this->getMockBuilder(ServiceLocatorInterface::class)
+            ->getMock();
```
webimpress: you can use here just `$this->getMock(ServiceLocatorInterface::class)`

tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
webimpress: Not sure if these tests should have `expects` at all. This is not part of the scope of the test, so I think it should be removed (this is why I was suggesting `prophesies`) - it doesn't matter how many times `->get('config')` / `-> getServiceLocator` / ...  will be called etc

tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
thomasvargiu: @webimpress It's deprecated since 5.4, so I prefer to start using `getMockBuilder`.

I was thinking to enable PHPUnit `^4.8 || ^5.2`. What do you think?

webimpress: @thomasvargiu Yes, it is deprecated, but there is method createMock in `PHPUnit 5.4`. I would prefer migrate to `PHPUnit 5.4` (as separate PR). It is no point of supporting two PHPUnit versions as it is DEV requirement :)

thomasvargiu: @webimpress I agree, it also requires to refactor all `getMock()` calls.
I changed the mock creation for simple mocks. I prefer to use the Mock Builder when I need to disable the constructor, it's more declarative and readable.

webimpress: @thomasvargiu agree :) Thanks for it !

tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
```diff
@@ -0,0 +1,102 @@
...
+        $pluginManager->expects($this->once())->method('getServiceLocator')->willReturn($serviceLocator);
+
+        $factory = new YumlControllerFactory();
+        $controller = $factory->createService($pluginManager);
+
+        $this->assertInstanceOf(YumlController::class, $controller);
+    }
+
+    /**
+     * @expectedException \Zend\ServiceManager\Exception\ServiceNotFoundException
```
webimpress: I would prefer to use methods instead of these annotations. Why? Because then we can set expected method JUST before the call which should throw an exception. It could be changed later on in separate PR (changing to PHPUnit 5.4?)

tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
```diff
@@ -0,0 +1,101 @@
...
+
+        $serviceLocator->expects($this->once())->method('get')->with('config')->willReturn($config);
+        $pluginManager->expects($this->once())->method('getServiceLocator')->willReturn($serviceLocator);
+
+        $factory = new YumlControllerFactory();
+
+        $this->setExpectedException(\Zend\ServiceManager\Exception\ServiceNotFoundException::class);
+        $controller = $factory->createService($pluginManager);
+
+        $this->assertInstanceOf(YumlController::class, $controller);
```
webimpress: if we expect exception assertion is useless here

tests/DoctrineORMModuleTest/Yuml/YumlControllerFactoryTest.php
```diff
@@ -0,0 +1,101 @@
...
```
webimpress: as above

src/DoctrineORMModule/Yuml/YumlControllerFactory.php
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public function __invoke(ContainerInterface $container, $requestedName, array $options = null)
+    {
+        $config = $container->get('config');
+
+        if (! isset($config['zenddevelopertools']['toolbar']['enabled'])
+            || !$config['zenddevelopertools']['toolbar']['enabled']
+        ) {
+            throw new ServiceNotFoundException(
+                'Service DoctrineORMModule\\Yuml\\YumlController could not be found'
```
webimpress: I would prefer to use here `sprintf` + `::class`, what do you think?

src/DoctrineORMModule/Yuml/YumlControllerFactory.php
```diff
@@ -0,0 +1,76 @@
...
+     * Create an object
+     *
+     * @param  ContainerInterface $container
+     * @param  string             $requestedName
+     * @param  null|array         $options
+     * @return YumlController
+     * @throws ServiceNotFoundException if unable to resolve the service.
+     * @throws ServiceNotCreatedException if an exception is raised when
+     *     creating a service.
+     * @throws ContainerException if any other error occurs
```
webimpress: hm... I can see here only `ServiceNotFoundException`, do we need all of it?

thomasvargiu:
I made some CS fix and squashed commits

webimpress:
Looks good to me now! Thanks @thomasvargiu!

@gianarb can you merge it please?

gianarb:
I am waiting Travis :)

webimpress:
@gianarb all pass !

gianarb:
Thanks @thomasvargiu for your effort and @webimpress for your review :)

PR #496, opened 2016-08-10
Fix: Missing view helper "zendDeveloperToolsTime"

MatthiasKuehneEllerhold:
Fixes the error 'A plugin by the name "zendDeveloperToolsTime" was not found in the plugin manager'

### DIFF
--- a/view/zend-developer-tools/toolbar/doctrine-orm-queries.phtml
+++ b/view/zend-developer-tools/toolbar/doctrine-orm-queries.phtml
@@ -5,7 +5,7 @@
         <span class="zdt-toolbar-info">
             <?php echo $collector->getQueryCount(); ?>
             queries in
-            <?php echo $this->zendDeveloperToolsTime($collector->getQueryTime()); ?>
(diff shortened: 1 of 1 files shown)


hisamu:
Having the same issue here, this change fixed it.

gianarb:
Thanks!

exptom:
@gianarb Is it possible to get this merged soon?

gianarb:
This fix broke something with people that are using ZF2?

I think yes but I am not sure.. Anyone can try with ZF2? :) 
Thanks a lot

exptom:
@gianarb There shouldn't be a problem with ZF2. The service names are case-insensitive in ZF2 but they are case sensitive in ZF3. 

As you can see the view helper is registered with the uppercase 'Z': https://github.com/zendframework/ZendDeveloperTools/blob/master/src/Module.php#L132

gianarb:
Thanks @exptom  and thanks @MatthiasKuehneEllerhold for this PR

MatthiasKuehneEllerhold:
You're welcome and thanks for merging!

PR #497, opened 2016-08-10
PHPUnit update

kokspflanze:
changed `getMock` to `getMockBuilder` for PHPUnit 5.4+
changed `setExpectedException` to `expectException` it is already deprecated

changed arraysyntax in unittests
updated composer.json to min PHPUnit 5.5.0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ vendor/
 phpunit.xml
 .project/
 nbproject/
+composer.phar
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
         "zendframework/zend-servicemanager": "^2.7.6 || ^3.1.1"
     },
     "require-dev":       {
-        "phpunit/phpunit":                    "^4.8",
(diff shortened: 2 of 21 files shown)


webimpress:
@kokspflanze `getMock` is deprecated but you can use `createMock` instead in PHPUnit 5. It is shorter than `...->getMockBuilder(...)->getMock()`.

gianarb:
Thanks @kokspflanze 

@Ocramius do we have a sort of standard? will we use createMock or getMockBuilder? 

👍 getMockBuilder
💯 createMock

Reaction please! @webimpress :)

webimpress:
For me both are fine. Personally I'd prefer shorter version when possible. I would also go with `::class` notation, but it could be done later on in separate PR. It will be much nicer!

kokspflanze:
about the `::class` i will create a new PR after my vacation

gianarb:
Thanks a lot @kokspflanze

PR #504, opened 2016-09-26
Removed direct interface implementations

webimpress:
These will be phased out of ModuleManager and pushed to the appropriate components in the near future.

Module Manager checks if interface is implemented or method exists.
This direct implementation of interfaces is removed in all ZF and Apigility modules so it should be removed also here for future compatibility. **No BC Break!**


Ocramius:
Same remarks as https://github.com/doctrine/DoctrineModule/pull/574

PR #505, opened 2016-10-03
Require stable version of DoctrineModule 1.2.0 instead of dev-master

webimpress:
Thanks @gianarb for DoctrineModule 1.2.0 release with ZF3 compatibility ! 👍

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
     },
     "require":           {
         "php":                               "^5.6 || ^7.0",
-        "doctrine/doctrine-module":          "dev-master",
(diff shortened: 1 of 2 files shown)


gianarb:
I am not sure about this.. I think that we can work with `>=1.0` or something similar.. 
What do you think @webimpress ?

webimpress:
@gianarb Probably we can. Depends on policy you have in Doctrine. In Zend modules we updated all packages to require last stable version. If somebody can update this library to the latest version it means his code is also compatible with DoctrineModule 1.2.0 and it should be also updated IMO.

gianarb:
Right :) Thanks

PR #507, opened 2016-10-16
General code fixes

webimpress:
- updated PHP CodeSniffer to version 2.7
- added phpcs.xml with defined CS rules
- fixed code to pass all CS checks
- short array syntax in docs
- composer scripts to run unit and cs checks
- updated travis configuration to use composer scripts

ping @gianarb

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,9 +68,9 @@ install:
   - composer show
 
 script:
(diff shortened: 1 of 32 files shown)


TomHAnderson:
Merged

PR #520, opened 2017-01-07
Added ConfigProvider in line with ZF components

asgrim:
Many ZF components now use a `ConfigProvider` to assist in config merging and allow components to be used e.g. in Expressive.

I took the strategy of least interference with existing config loading, so I did not move config here, nor make `Module.php` load config using the `ConfigProvider`. If desired, however, I can add this functionality into this PR. Feedback welcome, thanks!


TomHAnderson:
Will you share another implementation of this pattern for reference?

asgrim:
https://github.com/zendframework/zend-form/blob/master/src/ConfigProvider.php

PR #521, opened 2017-01-16
DBALConnectionFactory opens db connection, even if its not necessary

IceShack:
Hi guys,

We sometimes got the issue that our app can't connect to our database.
Right now we just get a "ServiceNotCreatedException", because the DBALConnectionFactory opens a db connection while creating the connection services.

This fix only opens a db connection, if needed.
That way we can catch the ConnectionException in our controller/business logic and return a proper response.

### DIFF
--- a/src/DoctrineORMModule/Service/DBALConnectionFactory.php
+++ b/src/DoctrineORMModule/Service/DBALConnectionFactory.php
@@ -60,16 +60,12 @@ class DBALConnectionFactory extends AbstractFactory
         $eventManager  = $container->get($options->getEventManager());
 
         $connection = DriverManager::getConnection($params, $configuration, $eventManager);
+        foreach ($options->getDoctrineTypeMappings() as $dbType => $doctrineType) {
+            $connection->getDatabasePlatform()->registerDoctrineTypeMapping($dbType, $doctrineType);
+        }
 
-        if (count($options->getDoctrineTypeMappings()) > 0 || $options->getDoctrineCommentedTypes() > 0) {
-            $platform = $connection->getDatabasePlatform();
-            foreach ($options->getDoctrineTypeMappings() as $dbType => $doctrineType) {
-                $platform->registerDoctrineTypeMapping($dbType, $doctrineType);
-            }
-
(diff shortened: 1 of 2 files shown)


src/DoctrineORMModule/Service/DBALConnectionFactory.php
```diff
@@ -60,13 +60,16 @@ public function __invoke(ContainerInterface $container, $requestedName, array $o
...
 
-        foreach ($options->getDoctrineCommentedTypes() as $type) {
-            $platform->markDoctrineTypeCommented(Type::getType($type));
+        if (count($options->getDoctrineTypeMappings()) > 0 || $options->getDoctrineCommentedTypes() > 0) {
+            $platform = $connection->getDatabasePlatform();
+            foreach ($options->getDoctrineTypeMappings() as $dbType => $doctrineType) {
+                $platform->registerDoctrineTypeMapping($dbType, $doctrineType);
+            }
+
+            foreach ($options->getDoctrineCommentedTypes() as $type) {
```
Ocramius: This is a hidden addition that isn't tested

src/DoctrineORMModule/Service/DBALConnectionFactory.php
```diff
@@ -60,13 +60,16 @@ public function __invoke(ContainerInterface $container, $requestedName, array $o
         $eventManager  = $container->get($options->getEventManager());
 
         $connection = DriverManager::getConnection($params, $configuration, $eventManager);
-        $platform = $connection->getDatabasePlatform();
-        foreach ($options->getDoctrineTypeMappings() as $dbType => $doctrineType) {
-            $platform->registerDoctrineTypeMapping($dbType, $doctrineType);
```
Ocramius: A simpler fix is to simply this change this line to `$connection->getDatabasePlatform()->registerDoctrineTypeMapping($dbType, $doctrineType);`

src/DoctrineORMModule/Service/DBALConnectionFactory.php
IceShack: That was there all along, i only added the if.

Ocramius:
LGTM 👍

PR #527, opened 2017-08-16
Support for FilterSchemaAssetsExpression

fbrinker:
Added schema_filter (setFilterSchemaAssetsExpression) support for Zend in the configuration file like in symfony, described here: https://symfony.com/doc/current/bundles/DoctrineMigrationsBundle/index.html#manual-tables

### DIFF
--- a/src/DoctrineORMModule/Options/Configuration.php
+++ b/src/DoctrineORMModule/Options/Configuration.php
@@ -201,7 +201,7 @@ class Configuration extends DBALConfiguration
      * @var SecondLevelCacheConfiguration|null
      */
     protected $secondLevelCache;
-    
+
     /**
      * Configuration for schema filter
      *
--- a/src/DoctrineORMModule/Options/Configuration.php
+++ b/src/DoctrineORMModule/Options/Configuration.php
@@ -35,7 +35,7 @@ class Configuration extends DBALConfiguration
      * @var string
      */
     protected $queryCache = 'array';
-    
+
     /**
      * Set the cache key for the result cache. Cache key
      * is assembled as "doctrine.cache.{key}" and pulled from
--- a/src/DoctrineORMModule/Options/Configuration.php
+++ b/src/DoctrineORMModule/Options/Configuration.php
@@ -203,11 +203,11 @@ class Configuration extends DBALConfiguration
     protected $secondLevelCache;
 
     /**
-     * Configuration for schema filter
+     * Configuration option for the filter schema expression
      *
      * @var string|null
      */
-    protected $schemaFilter;
+    protected $filterSchemaAssetsExpression;
 
     /**
      * @param  array $datetimeFunctions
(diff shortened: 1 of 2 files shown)


src/DoctrineORMModule/Options/Configuration.php
```diff
@@ -35,7 +35,7 @@ class Configuration extends DBALConfiguration
      * @var string
      */
     protected $queryCache = 'array';
-
+
```
TomHAnderson: Please remove this unnecessary white space too.

src/DoctrineORMModule/Service/ConfigurationFactory.php
```diff
@@ -152,6 +152,10 @@ public function __invoke(ContainerInterface $container, $requestedName, array $o
             $config->setSecondLevelCacheEnabled();
             $config->setSecondLevelCacheConfiguration($cacheConfiguration);
         }
+        
+        if($schemaFilter = $options->getSchemaFilter()) {
```
TomHAnderson: Why are you using `getSchemaFilter` when this passes to the function `setFilterSchemaAssetsExpression`?  It seems to me we should keep the `setFilterSchemaAssetsExpression` name for the `Configuration`.

fbrinker: We could, but in my opinion it is easier for everyone if we use the same naming as in the symfony repository. There is the config option named schema_filter and so it translates to set/getSchemaFilter. So it's sort of "compatible"..?

Is your suggestion naming the option filter_schema, or filter_schema_assets_expression? The last one would be a little bit too much?

TomHAnderson: I believe in calling an elephant an elephant and not an ele.  This is why I asked for your reasoning for shortening the name.  

The existing documentation for the SchemaTool is:
https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/DropCommand.php#L73
and is says `$config->setFilterSchemaAssetsExpression($regexp);`

This is correct and this is what you are doing.  Having read this and seeing that it is a regular expression and filters the schema (of not just table assets) but mostly because the term is already used in the SchemaTool, I would like this configuration key to be `filter_schema_assets_expression` and not the short hand of `filter_schema` which sounds like a boolean to me.

TomHAnderson: p.s. I'm a proponent of constant diction in an application.  So 
```
$config->setFilterSchemaAssetsExpression($schemaFilter);
```
breaks constant diction because the same thing is being called two things here.

fbrinker: I agree with you. I just thought it could be nice to adopt the same option name they've used in the symfony bundle :)

TomHAnderson:
Follup up with https://github.com/doctrine/DoctrineORMModule/pull/529

TomHAnderson:
Will be in next release after https://github.com/doctrine/DoctrineORMModule/pull/528 is merged.

TomHAnderson:
Released in 1.1.4

PR #532, opened 2017-08-20
drop hhvm from travis

kokspflanze:
like in [zf-blog](https://framework.zend.com/blog/2017-06-06-zf-php-7-1.html) written removed hhvm

also the current hhvm tests go to a vm error, so it was not realy testet

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,18 +54,8 @@ matrix:
     - php: nightly
       env:
         - DEPS=latest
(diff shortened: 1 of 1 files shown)


TomHAnderson:
@Ocramius I'm behind this 100%.  Can I get a second opinion?

Ocramius:
@TomHAnderson this is fine :+1:

PR #534, opened 2017-09-08
Allow retrieving platform from container

thomasvargiu:
This PR allows to retrieve platform from container.

The `platform` param must be an instance, but it's impossible to set an instance in the configuration (if we want to cache it). So, if we provide a string, we should check if can be retrieved from the service container.

### DIFF
--- a/tests/DoctrineORMModuleTest/Service/DBALConnectionFactoryTest.php
+++ b/tests/DoctrineORMModuleTest/Service/DBALConnectionFactoryTest.php
@@ -19,6 +19,7 @@
 
 namespace DoctrineORMModuleTest\Service;
 
+use Doctrine\DBAL\Platforms\AbstractPlatform;
 use PHPUnit\Framework\TestCase;
 use DoctrineORMModuleTest\Assets\Types\MoneyType;
 use DoctrineORMModule\Service\DBALConnectionFactory;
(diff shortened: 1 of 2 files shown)


tests/DoctrineORMModuleTest/Service/DBALConnectionFactoryTest.php
```diff
@@ -154,4 +155,36 @@ public function testDoctrineAddCustomCommentedType()
...
         $this->assertTrue($platform->isCommentedDoctrineType($type));
     }
+
+    public function testGettingPlatformFromContainer()
+    {
+        $config = [
+            'doctrine' => [
+                'connection' => [
+                    'orm_default' => [
+                        'driverClass'   => \Doctrine\DBAL\Driver\PDOSqlite\Driver::class,
```
TomHAnderson: For all instances where you give the fully qualified namespace please replace with a `use` at the top of the script and just the class name in the test.

tests/DoctrineORMModuleTest/Service/DBALConnectionFactoryTest.php
```diff
@@ -154,4 +155,36 @@ public function testDoctrineAddCustomCommentedType()
...
+                        ],
+                    ]
+                ],
+            ],
+        ];
+        $configurationMock = $this->getMockBuilder(\Doctrine\ORM\Configuration::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $platformMock = $this->getMockBuilder(AbstractPlatform::class)
```
TomHAnderson: This class use is correct

tests/DoctrineORMModuleTest/Service/DBALConnectionFactoryTest.php
```diff
@@ -154,4 +155,36 @@ public function testDoctrineAddCustomCommentedType()
...
+                    'orm_default' => [
+                        'driverClass'   => \Doctrine\DBAL\Driver\PDOSqlite\Driver::class,
+                        'params' => [
+                            'platform' => 'platform_service',
+                        ],
+                    ]
+                ],
+            ],
+        ];
+        $configurationMock = $this->getMockBuilder(\Doctrine\ORM\Configuration::class)
```
TomHAnderson: This class use is incorrect

thomasvargiu:
@TomHAnderson done

PR #538, opened 2017-11-13
Uses listener class to configure console application

neeckeloo:
Hi!

This PR is not adding any features, it's merely the start of a refactoring in order to split responsibilities in some of the existing classes. The goal is to get out the logic that initializes the console application from the `DoctrineORMModule\Module` class.

 * the listener of `loadCli.post` event, that aims to initialize console application, is separated in it's own file and class:`DoctrineORMModule\Listener\PostCliLoadListener`
 * `DoctrineORMModule\CliConfigurator` class is used to configure console application in the previous listener. This class has been splitted into several methods that could lead to the creation of other classes to improve separation of concerns.
* consequently, unit tests have been updated and improved

### DIFF
--- a/config/module.config.php
+++ b/config/module.config.php
@@ -22,8 +22,10 @@ use Doctrine\DBAL\Tools\Console;
 use Doctrine\ORM\Tools\Console\Command;
 use DoctrineModule\Form\Element;
 use DoctrineModule\Stdlib\Hydrator\DoctrineObject;
+use DoctrineORMModule\Listener\PostCliLoadListener;
 use DoctrineORMModule\Service;
 use DoctrineORMModule\Yuml;
+use Zend\ServiceManager\Factory\InvokableFactory;
 
 return [
     'doctrine' => [
@@ -195,6 +197,7 @@ return [
 
     'service_manager' => [
         'factories' => [
+            PostCliLoadListener::class => InvokableFactory::class,
             'Doctrine\ORM\EntityManager' => Service\EntityManagerAliasCompatFactory::class,
         ],
         'invokables' => [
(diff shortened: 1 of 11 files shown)


TomHAnderson:
Prefixed name to [wip].   Rename after this is no longer a work in progress.

TomHAnderson:
Is this a good time to rewrite the console commands to use zendframework/zend-mvc-console ?

MatthiasKuehneEllerhold:
`zendframework/zend-mvc-console` is deprecated, see: https://zendframework.github.io/zend-mvc-console/intro/#deprecated.

neeckeloo:
I'm not sure it's really relevant. Indeed, the Doctrine commands used in `DoctrineORMModule` come from `doctrine/dbal`, `doctrine/orm` and `doctrine/migrations` packages which use `symfony/console`. In this context, I do not see how we could use `zfcampus/zf-console` which is recommanded over `zendframework/zend-mvc-console`.

neeckeloo:
I consider this pull request as ready to be merged because the goal of getting the console initialization logic out of the module class has been achieved. However, the `DoctrineORMModule\CliConfigurator` class can still be improved in the future.

TomHAnderson:
@tdutrion do you have any input on this PR; can it be merged before your migrations work or as part of it?

tdutrion:
It can definitely be merged before, it just improve the whole code quality by splitting responsibilities into different blocks.
It basically has no impact at all on the PR I will submit.

TomHAnderson:
@neeckeloo not a line of code would I change.  Your style is the same as mine as this PR is concerned.  Very happy to have these clean changes in.

I'll create a release along with @tdutrion migrations changes since this isn't a priority.

PR #539, opened 2017-11-13
[wip] Refactor migration docs for upcoming multiple object manager support

TomHAnderson:


TomHAnderson:
Abandoned.  I think this should multiple migrations should be in their own repository like data fixtures https://github.com/API-Skeletons/zf-doctrine-data-fixture

PR #543, opened 2017-11-17
Bump minimum PHP version to 7.1.0

neeckeloo:


composer.json
```diff
@@ -37,7 +37,7 @@
         }
     },
     "require":           {
-        "php":                               "^5.6 || ^7.0",
+        "php":                               "^7.0",
```
alcaeus: Active support for PHP 7.0 drops in 15 days, so I'd suggest to require 7.1 by default.

kokspflanze:
than you can also change phpunit in composer.json to min version 6

neeckeloo:
Do you think it would be relevant to add scalar type and return type declarations in this PR?

kokspflanze:
@neeckeloo dont think so, it should be in a other pr with much testing =)

composer.json
```diff
@@ -37,7 +37,7 @@
         }
     },
     "require":           {
-        "php":                               "^5.6 || ^7.0",
+        "php":                               "^7.1",
         "doctrine/doctrine-module":          "^1.2",
         "doctrine/orm":                      ">=2.5,<2.7",
         "doctrine/dbal":                     ">=2.4,<2.7",
```
carusogabriel: [DBAL's `2.6` version](https://github.com/doctrine/dbal/releases/tag/v2.6.0) is optimized to PHP 7.1 :+1:

composer.json
```diff
@@ -37,7 +37,7 @@
         }
     },
     "require":           {
-        "php":                               "^5.6 || ^7.0",
+        "php":                               "^7.1",
         "doctrine/doctrine-module":          "^1.2",
         "doctrine/orm":                      ">=2.5,<2.7",
```
carusogabriel: [ORM's `2.6` version](https://github.com/doctrine/doctrine2/releases/tag/v2.6.0) is optimized to PHP 7.1 :+1:

composer.json
```diff
@@ -48,7 +48,7 @@
         "zendframework/zend-servicemanager": "^2.7.6 || ^3.1"
     },
     "require-dev":       {
-        "phpunit/phpunit":                    "^5.7.17 || ^6.2.1",
+        "phpunit/phpunit": "^6.4.4",
```
carusogabriel: We could use `PHPUnit 6.5` here.

composer.json
Ocramius: Yeah, this should be `^2.6.0`

composer.json
Ocramius: Yeah, this should be `^2.6.0`

TomHAnderson:
@Ocramius do I need to update the major version number when I merge this?

alcaeus:
Not @Ocramius, but no: people not on PHP 7.1 will simply not get an update. That is unless they have their environments messed up, in which case I'll happily refer them to http://www.doctrine-project.org/2017/07/25/php-7.1-requirement-and-composer.html.

composer.json
carusogabriel: PHPUnit 7 can be used here

PR #546, opened 2017-11-27
[DOCS] Fix the documentation for using multiple connections

MatthiasKuehneEllerhold:
In the documentation for using multiple connection is written that we should register 8 factories to make the new connection, Entity-Manager, etc ... work. See https://github.com/doctrine/DoctrineORMModule/blob/master/docs/configuration.md#how-to-use-two-connections
This worked well in dev, but fails everywhere else! 
```
Error: Call to undefined method DoctrineORMModule\Service\DBALConnectionFactory::__set_state() in /x/data/module-config-cache.php:187
Stack trace:
#0 /x/vendor/zendframework/zend-modulemanager/src/Listener/ConfigListener.php(392): include()
#1 /x/vendor/zendframework/zend-modulemanager/src/Listener/ConfigListener.php(70): Zend\ModuleManager\Listener\ConfigListener->getCachedConfig()
#2 /x/vendor/zendframework/zend-modulemanager/src/Listener/DefaultListenerAggregate.php(112): Zend\ModuleManager\Listener\ConfigListener->__construct(Object(Zend\ModuleManager\Listener\ListenerOptions))
#3 /x/vendor/zendframework/zend-modulemanager/src/Listener/DefaultListenerAggregate.php(42): Zend\ModuleManager\Listener\DefaultListenerAggregate->getConfigListener()
#4 /x/vendor/zendframework/zend-mvc/src/Service/ModuleManagerFactory.php(77): Zend\ModuleManager\Listener\DefaultListenerAggregate->attach(Object(Zend\EventManager\EventManager))
#5 /x/vendor/zendframework/zend-servicemanager/src/ServiceManager.php(758): Zend\Mvc\Service\ModuleManagerFactory->__invoke(Object(Zend\ServiceManager\ServiceManager), 'ModuleManager', NULL)
#6 /x/vendor/zendframework/zend-servicemanager/src/ServiceManager.php(200): Zend\ServiceManager\ServiceManager->doCreate('ModuleManager')
#7 /x/vendor/zendframework/zend-mvc/src/Application.php(264): Zend\ServiceManager\ServiceManager->get('ModuleManager')
#8 /x/bin/doctrine-x.php(14): Zend\Mvc\Application::init(Array)
#9 {main}
```
Because the config cache is enabled (except in dev) and these factories do not implement the necessary method. This means we cant use these factories directly. There is an AbstractFactory that manages all that and is cache-able: `DoctrineModule\ServiceFactory\AbstractDoctrineServiceFactory`. 

So this PR removes every mention of these factories so we can cache this config.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -114,8 +114,9 @@ return [
     'doctrine' => [
         'connection' => [
             'orm_crawler' => [
-                'driverClass' => 'Doctrine\DBAL\Driver\PDOMySql\Driver',
-                'params' => [
+                'driverClass'   => 'Doctrine\DBAL\Driver\PDOMySql\Driver',
+                'configuration' => 'orm_crawler',
+                'params'        => [
                     'host'     => 'localhost',
                     'port'     => '3306',
                     'user'     => 'root',
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -180,28 +180,16 @@ return [
 ];
 ```
 
-Module.php
-```php
-public function getServiceConfig()
-{
-    return [
-        'factories' => [
-            'doctrine.connection.orm_crawler'           => new \DoctrineORMModule\Service\DBALConnectionFactory('orm_crawler'),
-            'doctrine.configuration.orm_crawler'        => new \DoctrineORMModule\Service\ConfigurationFactory('orm_crawler'),
-            'doctrine.entitymanager.orm_crawler'        => new \DoctrineORMModule\Service\EntityManagerFactory('orm_crawler'),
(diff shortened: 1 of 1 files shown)


MatthiasKuehneEllerhold:
Debugging some weird edge-cases I noticed that my second DBAL connection didnt get the correct configuration. They were shared across all connections.

Debugging through it made clear that you have to set explicitly set the configuration in the connection for it to work properly. Added it to the docs.

@Ocramius Any news on this? Is this PR still relevant with the current-rewrite in progress?

TomHAnderson:
I too found the documentation lacking so I wrote my own.  Please reference my blog article and update your PR to reflect the correct way to create two connections:  http://blog.tomhanderson.com/2016/03/zf2-doctrine-configure-second-object.html

PR #554, opened 2018-03-25
allow to configure custom template

waahhhh:
allow to configure custom template

### DIFF
--- a/src/DoctrineORMModule/Service/MigrationsConfigurationFactory.php
+++ b/src/DoctrineORMModule/Service/MigrationsConfigurationFactory.php
@@ -48,6 +48,10 @@ class MigrationsConfigurationFactory extends AbstractFactory
             $configuration->setMigrationsColumnName($migrationsConfig['column']);
         }
 
+        if (method_exists($configuration, 'setCustomTemplate')) {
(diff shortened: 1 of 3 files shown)


waahhhh:
Hm... i used the wrong the default value... it has to be null instead of false. I'll correct it.

src/DoctrineORMModule/Service/MigrationsConfigurationFactory.php
```diff
@@ -48,6 +48,10 @@ public function __invoke(ContainerInterface $container, $requestedName, array $o
             $configuration->setMigrationsColumnName($migrationsConfig['column']);
         }
 
+        if (method_exists($configuration, 'setCustomTemplate')) {
+            $configuration->setCustomTemplate($migrationsConfig['custom_template']);
```
TomHAnderson: I think in order to cover all the bases you need to verify `custom_template` exists on the `$migrationsConfig` before setting it.

waahhhh: Your'e right! i added a check to ensure `custom_template` exists of the `$migrationsConfig` before accessing it.

PR #582, opened 2019-01-10
Added support for doctrine/migrations v2

### LINKED ISSUE #581: migrations 2.x is not compatible
when upgrading to doctrine/migrations 2.x all migrations commands are no longer working because of some namespace changes in version 2.x (removal of DBAL part in the namespace)

webimpress:
Resolves #581 

- changed all usage to new namespace
- created `autoload.php` file in `src` to create aliases to new classes in case v1 is used


TomHAnderson:
Should this line be removed?
https://github.com/doctrine/DoctrineORMModule/blob/master/tests/TestConfigurationV3.php#L8

Issue #567, opened 2018-04-16
Fatal Error in Module class at Version 1.1.8


DennisDobslaf:
https://github.com/doctrine/DoctrineORMModule/blob/a205cc69e35633ced3723629dc54f7edea677841/src/DoctrineORMModule/Module.php#L56

In my case this leads to a fatal error. 
Message:
"Fatal error: Uncaught Error: Call to undefined method Zend\ModuleManager\ModuleManager::getParam() in \vendor\doctrine\doctrine-orm-module\src\DoctrineORMModule\Module.php on line 56"

It seems, that $event->getTarget() is instanceof Zend\ModuleManager\ModuleManager, which has no getParam() method!

TomHAnderson:
Since releaseing 1.1.8 I've had a report that this line **does** work and your report that this line **does not** work.

Here's the line in question: https://github.com/doctrine/DoctrineORMModule/blob/release/1.1.8/src/DoctrineORMModule/Module.php#L56

This is triggered here:
https://github.com/zendframework/ZendDeveloperTools/blob/master/src/Module.php#L63

In the triggering the $event is passed a Zend\ModuleManager\ModuleEvent for 1.1.8 and 2.1.5 which both address this listener.

Please post your `composer show` result.

thomasvargiu:
I think could be a issue between different `zend-modulemanager` releases.

Btw we know that it worked in version `1.1.5`. [Here the difference](https://github.com/doctrine/DoctrineORMModule/compare/1.1.5...1.1.8#diff-91b9c94f2f305f9007116b4acd7be06bL55).

Now the event used for retrieving the ServiceManager is not the same as in release `1.1.5`.
We should change it like this:

```php
$manager
    ->getEventManager()
    ->attach(
        ProfilerEvent::EVENT_PROFILER_INIT,
        function () use ($manager) {
            $container = $manager->getEvent()->getTarget()->getParam('ServiceManager');
            $container->get('doctrine.sql_logger_collector.orm_default');
        }
    );
```

@DennisDobslaf we still need your `composer show` result to understand which module causes the issue.
@TomHAnderson (almost) every package used in `use` statement I think should be in the `composer.json` as requirement, including `zend-eventmanager`, `zend-modulemanager`, etc...

DennisDobslaf:
off course. here is the composer show result

`acelaya/zf2-acmailer                        v4.5.1                                      Mail sending module for Zend Framework 2
[...]

PR #585, opened 2019-03-01
 Making 1.1.8 work with zendframework/zend-developer-tools 1.2.1

bertrandgauthier:
Related to forgotten [doctrine/DoctrineORMModule#567](https://github.com/doctrine/DoctrineORMModule/issues/567).


TomHAnderson:
This module has moved on to Laminas which is at 2.0.  Is this still an issue?  I've reviewed #567 and can't confirm that this is a change we should make.

TomHAnderson:
No reply from original poster.  Closing.

PR #590, opened 2019-07-07
Fix output writer for doctrine-migrations v2 integration

webimpress:
When we use doctrine-migrations v2 we can't see output on the console on running commands, for example:
```bash
vendor/bin/doctrine-module migrations:migrate
```

This is because in v2 we do have dependency factory and the same instance of dependency is always used instead of creating new. In the migration configuration factory however we are creating another instance which is not used in dependency factory later on.


webimpress:
The unit test to show invalid behaviour before is quite complicated (must be some kind of integration test). The other PR I created should be helpful to see the issue: #589 (we can notice there on builds with latests deps, there is no output about migrations - please compare with build with lowest deps).

PR #593, opened 2019-11-15
Enable use savepoints configuration

thomasvargiu:
This feature allows to configure connection to use savepoints.

### DIFF
--- a/src/DoctrineORMModule/Options/DBALConnection.php
+++ b/src/DoctrineORMModule/Options/DBALConnection.php
@@ -74,6 +74,11 @@ class DBALConnection extends AbstractOptions
     protected $doctrineCommentedTypes = [];
 
     /**
+     * @var bool
+     */
+    protected $useSavepoints = false;
+
+    /**
      * @param string $configuration
      */
     public function setConfiguration($configuration)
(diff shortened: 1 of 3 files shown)


src/DoctrineORMModule/Options/DBALConnection.php
```diff
@@ -204,4 +209,20 @@ public function getWrapperClass()
...
+    {
+        return $this->useSavepoints;
+    }
+
+    /**
+     * @param bool $useSavepoints
+     */
+    public function setUseSavepoints($useSavepoints)
+    {
+        $this->useSavepoints = $useSavepoints;
```
TomHAnderson: Add return $this for a fluent interface.

src/DoctrineORMModule/Service/DBALConnectionFactory.php
```diff
@@ -59,6 +59,10 @@ public function __invoke(ContainerInterface $container, $requestedName, array $o
         foreach ($options->getDoctrineCommentedTypes() as $type) {
             $connection->getDatabasePlatform()->markDoctrineTypeCommented(Type::getType($type));
         }
+        
+        if ($options->useSavepoints()) {
+            $connection->setNestTransactionsWithSavepoints(true);
```
TomHAnderson: Because savepoints is defaulted to false you can drop the if statement here and just use
```
$connection->setNestTransactionsWithSavepoints($options->useSavepoints());
```

src/DoctrineORMModule/Options/DBALConnection.php
thomasvargiu: @TomHAnderson This class has no fluent interface at all

src/DoctrineORMModule/Service/DBALConnectionFactory.php
thomasvargiu: I can't do this, because the `setNestTransactionsWithSavepoints()` method in the connection class always checks if the platform supports it and throws an exception if it doesn't (even when passing `false`).

PR #602, opened 2020-01-12
[wip] Migrate to Laminas

TomHAnderson:


samsonasik:
why upgrade to php 7.4?

TomHAnderson:
re version 7.4 minimum; phpunit 8.5 is 7.3 min, proxy-manager minimum 7.4
https://github.com/Ocramius/ProxyManager/blob/master/composer.json

samsonasik:
most of laminas components itself still support php 7.1, I think migration to laminas should still support php 7.1 once, create a minor/major version for it so bug fix can be backported there, then create another major version for php 7.4.

TomHAnderson:
I'm open if you want to create a composer.json which works for the latest versions of the libraries I have listed and goes back to 7.1 I'll be sure to review it for this library.

kokspflanze:
i dont realy understand why `ocramius/proxy-manager` is now a requirement, it was never required before.
i just see that `doctrine/migrations` need it and this is an optional requirement.
About phpunit, thats update is needed but, its just a require-dev dependency so i think it can be also minor/major version after the change to laminas.

dont miss we also need the new version of `doctrine/doctrine-module`, the current version is for zend and not laminas.

My problem with php74 is, that not all current zend-modules support php74, (not all modules tests on php74). so the update will be realy hard, to jump to php74 and to laminas at the same time.

MadCat34:
Globally, Im' ok with if this PR is a migration from Zend to Laminas, it only should update the composer.json to replace zend-* dependencies to laminas  (same versions) and change all the Zend\ namespace.

It should nor change the minimum supported PHP version nor add dependencies

TomHAnderson:
@MadCat34 the change will update libraries and importantly will comply with the Doctrine phpcs.  PHPUnit will pass as well.  This library is behind the times and your idea to just update zend to laminas would overlook a prime opportunity to bring this library up to date.

Slamdunk:
> This library is behind the times and your idea to just update zend to laminas would overlook a prime opportunity to bring this library up to date.

Ok, but doing so in one step would be a tough move for many users. I see no downside in having a v2.2 release with Zend-to-Laminas rename, marking v2 branch as EOL, and immediately release a v3.0 version with all the upgrades you suggest

basz:
Unless I am overlooking something, I think I have to agree with madcaps/slamdunk's view. There are many libraries depending on this library that work fine with less then php74. Going the currently proposed route would force all those libraries to require php74. Is that what we want?

michalbundyra:
I guess we should target here also PHP 7.2+ as it is in [DoctrineModule 3.0](https://github.com/doctrine/DoctrineModule/releases/tag/3.0.0).

basz:
No issues with php72

TomHAnderson:
Closing in favor of #603

PR #607, opened 2020-02-17
phpcs and phpunit work with latest Doctrine coding standard

TomHAnderson:
This PR upgrades phpunit and phpcs and adds the Doctrine Coding Standard.

phpcs passes with the Doctrine standard.  

In making the many changes to make phpcs pass everything that was 
```
/**
 * @param array $something
```

became
```
/**
 * @param mixed[] $something
```

To research the origin of data for each array would be over taxing.

### DIFF
--- a/src/DoctrineORMModule/Service/DBALConfigurationFactory.php
+++ b/src/DoctrineORMModule/Service/DBALConfigurationFactory.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace DoctrineORMModule\Service;
 
 use Doctrine\DBAL\Configuration;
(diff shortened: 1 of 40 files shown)


src/DoctrineORMModule/Service/DBALConfigurationFactory.php
```diff
@@ -9,60 +11,55 @@
...
     }
 
     /**
      * {@inheritDoc}
      *
-     * @return \Doctrine\DBAL\Configuration
+     * @return Configuration
      */
-    public function __invoke(ContainerInterface $container, $requestedName, array $options = null)
+    public function __invoke(ContainerInterface $container, $requestedName, ?array $options = null)
```
greg0ire: I think it would make sense to use a return type declaration here and below, it's still compatible with the parent method, isn't it?

greg0ire: And above too, in all factories in fact.

TomHAnderson: This factory pattern is what the parent classes use too so I'm not inclined to try to improve it.

PR #613, opened 2020-03-07
Move unit tests to docker

TomHAnderson:


TomHAnderson:
This needs to be updated to latest and retested

TomHAnderson:
This module does not use anything but sqlite.  Docker is overkill at this time.

PR #614, opened 2020-03-08
changed laminas-hydrator doctrine-object

kokspflanze:
`DoctrineModule\Stdlib\Hydrator\DoctrineObject` to `Doctrine\Laminas\Hydrator\DoctrineObject`
and added `doctrine/doctrine-laminas-hydrator` in require-dev

im not 100% sure, if you want this change.

i just wonder about this class not found warning of my IDE.


TomHAnderson:
This already exists in doctrine-module which is required by this module

https://github.com/doctrine/DoctrineModule/blob/master/composer.json#L52

kokspflanze:
@TomHAnderson yes, but the class is still wrong in the configuration

TomHAnderson:
So there were two separate issues here.  If you want to put in a PR for the module change I'll see that it's done asap.  Else I'll do it today.

kokspflanze:
https://github.com/doctrine/DoctrineORMModule/pull/616

Issue #620, opened 2020-06-17
BC breaks with Migrations 3.0


PoetikDragon:
When attempting to run `./vendor/bin/doctrine-module migrations:migrate` from the command line, a fatal error occurs:

`PHP Fatal error:  Uncaught Error: Call to undefined method Doctrine\Migrations\Configuration\Configuration::getOutputWriter() in /var/www/vendor/doctrine/doctrine-orm-module/src/DoctrineORMModule/Service/MigrationsConfigurationFactory.php:39`

From here: https://github.com/doctrine/migrations/blob/master/UPGRADE.md#code-bc-breaks

- REMOVED: Method Doctrine\Migrations\Configuration\Configuration#getOutputWriter() was removed, get the Psr\Log\LoggerInterface service from Doctrine\Migrations\DependencyFactory

greg0ire:
v3 is not supported: https://github.com/doctrine/DoctrineORMModule/blob/792d0627cfbfb3eedc87bb078d2aafcfd678363b/composer.json#L55

But since it is an optional dependency, the constraint is not enforced. Maybe we could use [the `conflict` key](https://getcomposer.org/doc/04-schema.md#conflict) for this? Or at least, should have, it's a bit late now I'm afraid :sweat_smile:

PoetikDragon:
I realize it is not supported yet, was posting here so it didn't get forgotten about and so others who want to use it before its fully supported might find some work-arounds.

I wrote a custom integration that works. Unfortunately, I discovered that migrations 3.0 even *existing* in the same project causes all of the other doctrine-module cli commands to throw the same error, even if I am not calling a migrations command.

greg0ire:
> Maybe we could use the conflict key for this?

Any thoughts about this?

TomHAnderson:
I've added it in PR #622 

```
[Pro@~/Projects/doctrine/doctrine-orm-module (master)]$ composer require doctrine/migrations 3.0
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for doctrine/doctrine-orm-module dev-master -> satisfiable by doctrine/doctrine-orm-module[dev-master].
    - doctrine/migrations 3.0.0 conflicts with doctrine/doctrine-orm-module[dev-master].
    - Installation request for doctrine/migrations 3.0 -> satisfiable by doctrine/migrations[3.0.0].


[...]

PR #621, opened 2020-06-18
Only include migrations commands if compatible

PoetikDragon:
This is a temporary fix for the error message when using the CLI tools with doctrine migrations 3.0 installed. This will only load the migrations commands if they're compatible with DoctrineORMModule. It does not provide support for migrations 3.0.

https://github.com/doctrine/DoctrineORMModule/issues/620


TomHAnderson:
Fixed by adding `conflict` for 3.0 #622

PoetikDragon:
#622 is not a fix and this issue should not be closed.

The pull request actually fixes the problem, and allows you to run migrations commands as usual.

The "conflict" does not fix the problem, does not prevent installing both packages, and doesn't even give an error. It literally does nothing.

PR #624, opened 2020-06-25
Updated invalid namespace reference

pimjansen:
I noticed that the default used config uses an invalid reference towards doctrine itself. It is reference towards the Doctrine\Common where it does not exists anymore.

Ofcourse i am able to extend it however on default Laminas/Zend will do a config merge and therefor fails

### DIFF
--- a/config/module.config.php
+++ b/config/module.config.php
@@ -1,6 +1,6 @@
 <?php
 
-use Doctrine\Common\Persistence\Mapping\Driver\MappingDriverChain;
+use Doctrine\Persistence\Mapping\Driver\MappingDriverChain;
 use Doctrine\DBAL\Tools\Console;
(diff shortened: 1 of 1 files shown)


pimjansen:
I have no idea what the diff is actually. I see that the build fails with the fact that it expects it at the old location however `doctrine/persistence` extends it at the new place.

Or are we missing a dependency on `doctrine/persistence` 1.x instead?

greg0ire:
I did the namespace move in version 1.3, I recommend you require the latest 1.3, or even 2.0

pimjansen:
This is my mistake. I see the stability was set wrong so it jumped to the unreleased 2.8.x-dev instead of 2.7 which is still linked to DoctrineCommon though

PR #636, opened 2020-11-08
Support for Migrations 3

TomHAnderson:


TomHAnderson:
https://github.com/laminas/laminas-mvc-skeleton/issues/28 is blocking

greg0ire (approved):
The code looks good, some if not all commit look like they should be squashed into previous commits.

TomHAnderson:
> The code looks good, some if not all commit look like they should be squashed into previous commits.

Agreed. I'll rebase them into a single commit.

SenseException (approved):

svycka:
@TomHAnderson isn't this BC break? `composer update` will break everything. at least add conflict with doctrine migrations <3.0

TomHAnderson:
@svycka The 3.0.x version of this repo continues with legacy 1 & 2 support for migrations.
Version 3.1.x has migrations 3.  I think it smells like a BC break but is not.

in composer.json
"doctrine/doctrine-orm-module": "~3.0.0" will properly update to 3.0.8

"doctrine/doctrine-orm-module": "~3.0" will update to 3.1.0

"doctrine/doctrine-orm-module": "~3.0",
"doctrine/migrations": "^2.0" - this combination, which you are worried about, will resolve to doctrine-orm-module 3.0.8

I do not think this is a BC break.  For users not using migrations already it will/could upgrade to 3.1.0 and if they are using migrations it will resolve to 3.0.8

I've tested all of the above and will gladly hear your concerns.

svycka:
@TomHAnderson it seems you are right just updated my project and it resolved to 3.0.8. I guess the composer is intelligent enough to look for dev dependencies when resolving dependencies even with --no-dev I guess all good then.

Just not sure it is good to drop old migrations. I haven't migrated yet and don't know how hard it will be and what problems it can cause I hope will not be stuck like with `doctrine/doctrine-module:4.1` being not installable.

TomHAnderson:
@svycka I don't understand what you mean about dropping old migrations.  Are you trying to say migrations 1 and 2 will no longer be supported?  If so then know that is not the case because 3.0.x and 3.1.x are still maintained.   See https://github.com/doctrine/DoctrineORMModule/blob/3.1.x/.doctrine-project.json for details on the different branches of this project.

svycka:
As I understand 3.0.x will get only bugfixes I don't know if this is ok or bad just saying that when you drop or change one of the main features would be better to release a major version even if it is not a BC break as I will have to update composer.json anyway when will be ready for migrations 3.0. But that's just my thoughts I am not always thinking right :D

Anyway not a problem for me as I have planned upgrade migrations soon.

PR #637, opened 2020-11-09
Enabled multiple migration configurations

TomHAnderson:
As I learned the migrations code in preparation for moving to Migrations 3 I found the reason behind why it was only possible to create a migration for one configuration.  The object manager was hard coded because injecting the configuration was not a strait-forward affair.  I used a Helper injected into the console application's HelperSet to provide the configuration and voila multiple configurations are now supported.

I wanted to create this PR now, before Migrations 3, because I just don't think it will ever happen otherwise.

### DIFF
--- a/src/DoctrineORMModule/Console/Helper/MigrationsConfigurationHelper.php
+++ b/src/DoctrineORMModule/Console/Helper/MigrationsConfigurationHelper.php
@@ -49,14 +49,9 @@ class MigrationsConfigurationHelper implements
 
     public function getMigrationConfig(InputInterface $input): Configuration
     {
-        try {
-            $objectManagerAlias = $input->getOption('object-manager');
-        } catch (Throwable $e) {
-            $objectManagerAlias = 'doctrine.entitymanager.orm_default';
-        }
+        $objectManagerAlias = $input->getOption('object-manager') ?: 'doctrine.entitymanager.orm_default';
+        $objectManagerName  = substr($objectManagerAlias, strrpos($objectManagerAlias, '.'));
 
-        $objectManagerName = substr($objectManagerAlias, strrpos($objectManagerAlias, '.') + 1);
-
-        return $this->container->get('doctrine.migrations_configuration.' . $objectManagerName);
+        return $this->container->get('doctrine.migrations_configuration' . $objectManagerName);
     }
 }
--- a/src/DoctrineORMModule/Console/Helper/MigrationsConfigurationHelper.php
+++ b/src/DoctrineORMModule/Console/Helper/MigrationsConfigurationHelper.php
@@ -29,7 +29,7 @@ class MigrationsConfigurationHelper implements
         $this->container = $container;
     }
 
-    public function setHelperSet(?HelperSet $helperSet = null): ConfigurationHelper
+    public function setHelperSet(?HelperSet $helperSet = null): MigrationsConfigurationHelper
     {
         $this->helperSet = $helperSet;
 
--- a/src/DoctrineORMModule/Console/Helper/MigrationsConfigurationHelper.php
+++ b/src/DoctrineORMModule/Console/Helper/MigrationsConfigurationHelper.php
@@ -10,10 +10,8 @@ use Interop\Container\ContainerInterface;
 use Symfony\Component\Console\Helper\HelperInterface;
 use Symfony\Component\Console\Helper\HelperSet;
 use Symfony\Component\Console\Input\InputInterface;
-use Throwable;
 
-use function strrpos;
-use function substr;
+use function preg_match;
 
 class MigrationsConfigurationHelper implements
     HelperInterface,
--- a/src/DoctrineORMModule/Console/Helper/MigrationsConfigurationHelper.php
+++ b/src/DoctrineORMModule/Console/Helper/MigrationsConfigurationHelper.php
@@ -14,7 +14,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use function strrpos;
 use function substr;
 
-class ConfigurationHelper implements
+class MigrationsConfigurationHelper implements
     HelperInterface,
     ConfigurationHelperInterface
 {
--- a/src/DoctrineORMModule/Console/Helper/MigrationsConfigurationHelper.php
+++ b/src/DoctrineORMModule/Console/Helper/MigrationsConfigurationHelper.php
@@ -10,6 +10,7 @@ use Interop\Container\ContainerInterface;
 use Symfony\Component\Console\Helper\HelperInterface;
 use Symfony\Component\Console\Helper\HelperSet;
 use Symfony\Component\Console\Input\InputInterface;
+use Throwable;
 
 use function strrpos;
 use function substr;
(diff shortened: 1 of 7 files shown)


src/DoctrineORMModule/Console/Helper/MigrationsConfigurationHelper.php
```diff
@@ -0,0 +1,62 @@
...
+
+    public function getMigrationConfig(InputInterface $input): Configuration
+    {
+        try {
+            $objectManagerAlias = $input->getOption('object-manager');
+        } catch (Throwable $e) {
+            $objectManagerAlias = 'doctrine.entitymanager.orm_default';
+        }
+
+        $objectManagerName = substr($objectManagerAlias, strrpos($objectManagerAlias, '.') + 1);
```
TomHAnderson: Is there a standard way of fetching the object manager name?

src/DoctrineORMModule/Console/Helper/MigrationsConfigurationHelper.php
```diff
@@ -0,0 +1,62 @@
...
+    public function getName(): string
+    {
+        return 'configuration';
+    }
+
+    public function getMigrationConfig(InputInterface $input): Configuration
+    {
+        try {
+            $objectManagerAlias = $input->getOption('object-manager');
+        } catch (Throwable $e) {
```
TomHAnderson: getOption will never fail on the commands so this needs to be simplified.

test/DoctrineORMModuleTest/Console/Helper/MigrationsConfigurationHelperTest.php
```diff
@@ -0,0 +1,88 @@
...
+        $request         = new Request([
+            'index.php',
+            '--object-manager=doctrine.entitymanager.orm_some_other_name',
+        ]);
+        $requestInput    = new RequestInput($request, $inputDefinition);
+        $helper          = new MigrationsConfigurationHelper($this->serviceLocator);
+
+        try {
+            $configuration = $helper->getMigrationConfig($requestInput);
+        } catch (ServiceNotFoundException $e) {
```
TomHAnderson: Change exception handling to annotation

test/DoctrineORMModuleTest/Console/Helper/MigrationsConfigurationHelperTest.php
```diff
@@ -0,0 +1,96 @@
...
+    {
+        $inputOption     = new InputOption('object-manager', [], InputOption::VALUE_REQUIRED);
+        $inputDefinition = new InputDefinition([$inputOption]);
+        $request         = new Request([]);
+        $requestInput    = new RequestInput($request, $inputDefinition);
+        $helper          = new MigrationsConfigurationHelper($this->serviceLocator);
+
+        $configuration = $helper->getMigrationConfig($requestInput);
+
+        $this->assertEquals(
```
SenseException:
```suggestion
        $this->assertSame(
```
We expect the same instance, right?

test/DoctrineORMModuleTest/Console/Helper/MigrationsConfigurationHelperTest.php
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * This test queries for a non-orm_default object manager
+     * and throws an exception when the configuration is not found
+     */
+    public function testGetNonDefaultMigrationConfig(): void
+    {
+        $this->expectException(ServiceNotFoundException::class);
+        $this->expectExceptionMessage(
```
SenseException, line 88: Please move those two expects-mrthods before the method that throws the expected exception.

test/DoctrineORMModuleTest/Console/Helper/MigrationsConfigurationHelperTest.php
```diff
@@ -0,0 +1,96 @@
...
+        $inputOption     = new InputOption('object-manager', [], InputOption::VALUE_REQUIRED);
+        $inputDefinition = new InputDefinition([$inputOption]);
+        $request         = new Request([
+            'index.php',
+            '--object-manager=doctrine.entitymanager.orm_some_other_name',
+        ]);
+        $requestInput    = new RequestInput($request, $inputDefinition);
+        $helper          = new MigrationsConfigurationHelper($this->serviceLocator);
+
+        $configuration = $helper->getMigrationConfig($requestInput);
```
SenseException:
```suggestion
        (new MigrationsConfigurationHelper($this->serviceLocator))->getMigrationConfig($requestInput);
```

or alternatively remove the unused `$configuration` variable.

SenseException (changes requested):
I'm missing tests in the helper where `$input->getOption('object-manager')` contains a value that matches (and doesn't matches) the regex and calls the expected service id.

Does `Module` have tests?

test/DoctrineORMModuleTest/Console/Helper/MigrationsConfigurationHelperTest.php
TomHAnderson: Moved to lines above call to getMigrationConfig

test/DoctrineORMModuleTest/Console/Helper/MigrationsConfigurationHelperTest.php
TomHAnderson: Used suggested code which does not assign the helper to a variable.

TomHAnderson:
> I'm missing tests in the helper where `$input->getOption('object-manager')` contains a value that matches (and doesn't matches) the regex and calls the expected service id.

Those tests have been added and the helper throws a RuntimeException if the alias is invalid.  However these tests are just to test the theoretical event that an object manager alias is invalid.  In practice an exception is thrown in the `CliConfigurator` if an invalid object manager alias exists.

 > Does `Module` have tests?

It does not.  The parts used in the `Module` have unit tests.

TomHAnderson:
A correction on whether Module.php is tested.  The testing framework creates the DoctrineORMModule within a module manager in `ServiceManagerFactory.php` using the config files at the root of the test directory.  All functions of the `Module.php` are tested and all the functionality in the closures is tested elsewhere.

test/DoctrineORMModuleTest/Console/Helper/MigrationsConfigurationHelperTest.php
```diff
@@ -0,0 +1,136 @@
...
+        (new MigrationsConfigurationHelper($this->serviceLocator))->getMigrationConfig($requestInput);
+    }
+
+    public function testValidObjectManagerAlias(): void
+    {
+        $inputOption     = new InputOption('object-manager', [], InputOption::VALUE_REQUIRED);
+        $inputDefinition = new InputDefinition([$inputOption]);
+        $request         = new Request([
+            'index.php',
+            '--object-manager=doctrine.entitymanager.orm_default',
```
SenseException: Can you please use a different id than the default manager? So that this test differs from the one for the default manager in case of regressions.

SenseException (changes requested):
Please rephrase the message of your last commit instead of the one-word `phpcs`.

test/DoctrineORMModuleTest/Console/Helper/MigrationsConfigurationHelperTest.php
TomHAnderson: Done and I removed the last commit and put those changes into the new commit for the second object manager.

greg0ire (approved):

SenseException (approved):

PR #640, opened 2020-11-16
Unit tests to exclude optional dependencies

TomHAnderson:
The multiple object manager migrations PR #637 overlooked cases where a developer has chosen not to include the optional `migrations` repository.  This PR duplicates all unit tests in the matrix to use a `composer` file which does not include any of the optional dependencies this module supports.  This did cause some problems and now all tests which tested optional dependency features will be skipped.

As part of this PR be sure to see the Travis build report.


test/DoctrineORMModuleTest/Console/Helper/MigrationsConfigurationHelperTest.php
```diff
@@ -46,6 +49,12 @@ public function setUp(): void
 
     public function testCreate(): void
     {
+        if (! class_exists(VersionCommand::class)) {
```
SenseException, line 47: Every test has this check in this class. Does it make sense to put it into `setUp()`?

SenseException (changes requested):
Any update in the composer.json also needs to be done in composer-no-optional-repos.json for any future changes in the dependencies. This might lead to issues in the future. What about doing a `composer remove` for the optional parts?

test/DoctrineORMModuleTest/Console/Helper/MigrationsConfigurationHelperTest.php
TomHAnderson: Done though `setUp()` is an odd place to call a function acting on a test.

TomHAnderson:
> Any update in the composer.json also needs to be done in composer-no-optional-repos.json for any future changes in the dependencies. This might lead to issues in the future. What about doing a `composer remove` for the optional parts?

I've done this and it is more simple this way.  Please see the `.travis.yml` file

test/DoctrineORMModuleTest/Console/Helper/MigrationsConfigurationHelperTest.php
SenseException: Only if really every test case in `MigrationsConfigurationHelperTest`  is affected.

SenseException (approved):

greg0ire (approved):

PR #643, opened 2020-12-06
Add support for Persistence 2.0 to 3.1.x

TomHAnderson:


composer.json
```diff
@@ -53,7 +53,7 @@
         "doctrine/dbal":                        "^2.10",
         "doctrine/doctrine-module":             "^4.0",
         "doctrine/orm":                         "^2.6.4",
-        "doctrine/persistence":                 "^1.3",
+        "doctrine/persistence":                 "^1.3.3|^2.0",
```
driehle: See https://github.com/doctrine/DoctrineORMModule/pull/644#discussion_r537365861
```suggestion
        "doctrine/persistence":                 "^1.3 || ^2.0",
```

driehle (changes requested):

snapshotpl (approved):

src/DoctrineORMModule/Form/Annotation/AnnotationBuilder.php
```diff
@@ -5,8 +5,8 @@
 namespace DoctrineORMModule\Form\Annotation;
 
 use ArrayObject;
-use Doctrine\Common\Persistence\Mapping\ClassMetadata;
-use Doctrine\Common\Persistence\ObjectManager;
+use Doctrine\Persistence\Mapping\ClassMetadata;
+use Doctrine\Persistence\ObjectManager;
```
greg0ire, line 9: Because of the `protected` function signature ligne 114, and since the class is not `final`, you will need to add a `class_exists(ClassMetadata::class);` to trigger the autoloading as per [my blog post](https://dev.to/greg0ire/how-to-deprecate-a-type-in-php-48cf)

test/DoctrineORMModuleTest/Assets/Fixture/TestFixture.php
```diff
@@ -3,7 +3,7 @@
 namespace DoctrineORMModuleTest\Assets\Fixture;
 
 use Doctrine\Common\DataFixtures\AbstractFixture;
-use Doctrine\Common\Persistence\ObjectManager;
+use Doctrine\Persistence\ObjectManager;
```
greg0ire, line 6: Same comment here because of `load()`

SenseException (approved):
:heavy_check_mark:  After the changes Grégoire suggested.

composer.json
```diff
@@ -53,7 +53,7 @@
         "doctrine/dbal":                        "^2.10",
         "doctrine/doctrine-module":             "^4.0",
         "doctrine/orm":                         "^2.6.4",
-        "doctrine/persistence":                 "^1.3",
+        "doctrine/persistence":                 "^1.3.3||^2.0",
```
svycka, line 56: do we really have to support v1 doctrine/persistance? https://github.com/doctrine/DoctrineModule/blob/4.1.x/composer.json#L54 or this for older doctrine-module versions? If not maybe just require `doctrine/doctrine-module:^4.1`

src/DoctrineORMModule/Form/Annotation/AnnotationBuilder.php
svycka: @greg0ire what if we only support `doctrine/persistance:^2` we could drop this hack? And `doctrine/doctrine-module:^4.1` already requires persistence:^2

composer.json
driehle: Though I think requiring `doctrine/doctrine-module:^4.1` would be reasonable I am unsure if this should be done in an `3.1.x` release. Loosing compatibility to `doctrine/persistence` within a bugfix release seems to break semantic versioning. So from my unserstanding you'd have to go to `3.2.x` when upgrading from `doctrine/doctrine-module:^4.0` to `^4.1`. Therefore, it might be better to keep it with `^1.3 || ^2.0` for now?

driehle:
@greg0ire @TomHAnderson I am just wondering if we can push this forward, since it currently prevents upgrading from orm 2.7 to 2.8.

TomHAnderson:
@driehle I'm actively working with all the involved parties to move this forward now.

src/DoctrineORMModule/Form/Annotation/AnnotationBuilder.php
greg0ire: @svycka in that case yes, the hack can be dropped.

TomHAnderson: I've dropped v1 persistence support https://github.com/doctrine/DoctrineORMModule/pull/643/commits/ed2bd4364b3df3a6ee4ab8490a2d1ffd4b07aea4

composer.json
TomHAnderson: v1 dropped in https://github.com/doctrine/DoctrineORMModule/pull/643/commits/ed2bd4364b3df3a6ee4ab8490a2d1ffd4b07aea4

TomHAnderson:
@SenseException The changes @greg0ire suggested were resolved to only supporting persistence v2 in this branch.  v1 is still supported in 3.0.  So this is now ready to merge.

composer.json
TomHAnderson: Agreed, driehle.  This PR is not the place to limit the persistence version to 2.0 only.

src/DoctrineORMModule/Form/Annotation/AnnotationBuilder.php
TomHAnderson: I won't be limiting this PR to persistence 2.0 only because this is not the right place for such a change and it would require a new minor version number.  I'll put the hack for `class_exists` in place.

TomHAnderson: @greg0ire your solution has been added

TomHAnderson:
@SenseException please consider approving this again.

composer.json
svycka: this whole PR is not a bugfix and will break semantic versioning, then we can release 3.2.0 no?

SenseException (approved):
@TomHAnderson Can you please address https://github.com/doctrine/DoctrineORMModule/pull/643#discussion_r537367080 in case `^1.3` can be used? Otherwise this PR is ready.

TomHAnderson:
> @TomHAnderson Can you please address [#643 (comment)](https://github.com/doctrine/DoctrineORMModule/pull/643#discussion_r537367080) in case `^1.3` can be used? Otherwise this PR is ready.

https://github.com/doctrine/DoctrineORMModule/pull/643/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R56

1.3 and 2.0 will be acceptable for this PR

greg0ire (approved):

driehle:
Thanks @TomHAnderson!

PR #644, opened 2020-12-06
Add support for persistence 2.0 to 3.0.x

TomHAnderson:


composer.json
```diff
@@ -53,7 +53,7 @@
         "doctrine/dbal":                        "^2.10",
         "doctrine/doctrine-module":             "^4.0",
         "doctrine/orm":                         "^2.6.4",
-        "doctrine/persistence":                 "^1.3",
+        "doctrine/persistence":                 "^1.3.3|^2.0",
```
driehle: @TomHAnderson Looks good to me, but I am unsure why you are requiring doctrine/persistance ^1.3.3. The forward compatibility has been added with 1.3.0 already, see https://github.com/doctrine/persistence/tree/1.3.0/lib/Doctrine. Besides, according to https://getcomposer.org/doc/articles/versions.md#version-range you should the a double pipe.

```suggestion
        "doctrine/persistence":                 "^1.3 || ^2.0",
```

driehle (changes requested):

SenseException:
Looks like the CS fixes are the last step before this can be approved.

greg0ire:
Also, the "Trigger Travis" commit can be dropped haha

greg0ire:
Ah this is missing some autoload calls, I will add them.

greg0ire:
It was just one call actually, done.

greg0ire (approved):

TomHAnderson:
> Looks like the CS fixes are the last step before this can be approved.

Yup.  Those are done now if you're ready to approve it

greg0ire:
@TomHAnderson please reset to 45efd9c, the `class_exists` is intended, and should not break CS (if it does that needs to be fixed somehow). Please read my lengthy commit message as to why this weird artefact is necessary.

TomHAnderson:
@greg0ire I restored your `class_exists` function and modified the phpcs.xml to allow for that usage.  @SenseException that fixes the cs check if I may now solicit your approval.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @TomHAnderson !

PR #647, opened 2020-12-12
Feature/GitHub actions

TomHAnderson:


greg0ire:
The build fails:

> Cannot load migrations from "data/DoctrineORMModule/Migrations" because it is not a valid directory

greg0ire (approved):

SenseException (commented):
@TomHAnderson Should the platform with the fixed PHP version be removed from composer.json?
```
"config": {
        "platform": {
            "php": "7.2.15"
        },
        "sort-packages": true
    },
```

SenseException (approved):

PR #650, opened 2021-01-13
PHP 8.0

TomHAnderson:


charma:
I'd really like php8 support. Thanks for your contribution! 

Why limited to patch-version 8.0.0?
Whats needed to merge this pull request?

TomHAnderson:
This was just a beginning to see what would be necessary to implement PHP 8.  If you'd like to submit a PR instead of this one you're welcome and encouraged to do so.

charma:
Not sure what else I'd change except of raising the dependency number in the composer.json and whereever you did, so it would result in the same PR.
I assume it's only the unit tests failing? I could have a look into that...

Slamdunk:
This is not as easy as it seems, see https://github.com/laminas/laminas-form/pull/75#issuecomment-769091279

TomHAnderson:
Closing in favor of #661

PR #654, opened 2021-01-26
Prevent E_USER_DEPRECATED error when creating RunSqlCommand.

bfoosness:
Version 2.11 of doctrine/dbal triggers an E_USER_DEPRECATED error when
instantiating a RunSqlCommand without a ConnectionProvider. Create a
factory for instantiating this command that passes a
SingleConnectionProvider if that class exists.

https://github.com/doctrine/dbal/blob/2.11.0/lib/Doctrine/DBAL/Tools/Console/Command/RunSqlCommand.php#L42

Would this solution be acceptable? Is there a better way to ensure backwards
compatibility since version 2.10 of doctrine/dbal does not support this?

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -274,8 +274,7 @@ Laminas Configuration
 How to Override RunSqlCommand Creation
 -------------------------
 
-The following Laminas configuration can be used to override the `creation
-<../src/DoctrineORMModule/Service/RunSqlCommandFactory.php>`_ of the
+The following Laminas configuration can be used to override the creation of the
 ``Doctrine\DBAL\Tools\Console\Command\RunSqlCommand`` instance used by this
 module.
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -227,7 +227,7 @@ How to Use Naming Strategy
 `Official documentation 
 <https://www.doctrine-project.org/projects/doctrine-orm/en/current/reference/namingstrategy.html>`__
 
-Zend Configuration
+Laminas Configuration
 
 .. code:: php
 
@@ -252,7 +252,7 @@ How to Use Quote Strategy
 `Official
 documentation <https://www.doctrine-project.org/projects/doctrine-orm/en/current/reference/basic-mapping.html#quoting-reserved-words>`__
 
-Zend Configuration
+Laminas Configuration
 
 .. code:: php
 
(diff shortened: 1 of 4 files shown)


SenseException:
@bfoosness I think a factory will make its usage much easier. Can you please add documentation about the factory and create tests to cover both code paths?

bfoosness:
> Can you please [...] create tests to cover both code paths?

For this functionality is there a better check than detecting the existence of `SingleConnectionProvider`? With that check the code paths are dependent on the version of DBAL, not this module's code, making it tricky to execute both of them. I could simulate the existence of `SingleConnectionProvider`, but that won't change the fact that the version of `RunSqlCommand` under test does not expect a connection.

The more I think about it, could this be a test coverage matrix problem? Does Codecov only validate [PHP 7.2.15](https://github.com/doctrine/DoctrineORMModule/blob/a5b649f0191ba61604674fe8cf6e7094edf3d5c3/composer.json#L40), which excludes later versions of DBAL? The other code path _would_ be covered in a later PHP version, but it's hard to imagine a scenario where they could both be covered in the same version (but I might be lacking imagination at the moment).

bfoosness:
>Can you please add documentation about the factory

Do you have a place in the docs in mind for this documentation and/or what you'd like documented? I could add a line to https://github.com/doctrine/DoctrineORMModule/tree/a5b649f0191ba61604674fe8cf6e7094edf3d5c3#registered-service-names, but this key (along with many others) has always been available for extension and wasn't listed there.

Or I could add a new "How to override RunSqlCommand creation" section to https://github.com/doctrine/DoctrineORMModule/blob/a5b649f0191ba61604674fe8cf6e7094edf3d5c3/docs/configuration.rst?

docs/configuration.rst
```diff
@@ -271,3 +271,21 @@ Zend Configuration
         ],
     ];
 
+How to Override RunSqlCommand Creation
+-------------------------
+
+The following Laminas configuration can be used to override the `creation
+<../src/DoctrineORMModule/Service/RunSqlCommandFactory.php>`_ of the
+``Doctrine\DBAL\Tools\Console\Command\RunSqlCommand`` instance used by this
+module.
```
bfoosness, line 279: I'm having a hard time getting the relative link working, but while I investigate that, do you think this will work?

TomHAnderson: I don't think trying to link to the source here is the best solution.  Not linking would be fine.

TomHAnderson:
I think this PR looks fine.  Please fix the docs and I'll merge it into the active branches.

docs/configuration.rst
bfoosness: Sure, will do.

TomHAnderson:
Version 3.2.x of this repository was created for enhancements to migrations.  I feel the 3.2.x should continue with enhancements and the 3.1.x branch should only receive bug fixes.  

That being said, are you OK with just putting this on the 3.2.x branch or do you think it belongs on the 3.1.x bug fix branch too?

bfoosness:
I just checked our codebase and we're clear to upgrade to 3.2.0, so I don't have a strong opinion either way 👍

TomHAnderson (approved):

SenseException (approved):
> The more I think about it, could this be a test coverage matrix problem?

I agree. I've checked the coverage and I agree that this is because of the current coverage matrix. I've also checked the CI workflow and dependencies of one matrix workflow should cover the other path of your code. Thank you @bfoosness.

TomHAnderson:
https://github.com/doctrine/DoctrineORMModule/releases/tag/3.2.1

bfoosness:
Thank you both for your help getting this merged.

PR #661, opened 2021-03-08
Feature/php8@3.2.2

TomHAnderson:


TomHAnderson:
This PR needs help.  A lot of annotation builder failures are preventing this from moving forward.

Slamdunk:
Because of https://github.com/laminas/laminas-form/pull/75#issuecomment-769091279

I suggest to drop annotation support and bump a new MAJOR

TomHAnderson:
@greg0ire @SenseException I've added you as reviewers because I'd like you opinion on dropping annotations support in Forms.  I would Love to drop the support because I don't like annotations as a rule, but I'll need a consensus among interested developers before I can do that here.

greg0ire:
:thinking: I don't use Laminas at all for now, so I don't have an opinion (or know what role annotations play in forms).

SenseException:
It looks like I'm going to drop laminas/laminas-form in one of my projects because it looks like php 8 support will take some time. But my project is not a module/library while DoctrineORMModule is and I don't know about the impact of dropping annotations support in forms.

Will you target a 3.3.x branch in case you drop the support? If laminas-form releases their php 8 support, will you reintroduce it in 3.4 or still in the 3.3 branch?

Slamdunk:
Hey there, I've released https://github.com/laminas/laminas-form/releases/tag/2.16.0 and run this PR against it, and the tests pass.

But we need to get rid of PHP 7.2 and

```json
    "config": {
        "platform": {
            "php": "7.2.15"
        },
    },
```

PR #670, opened 2021-10-14
Have `DoctrinePaginator` implement `JsonSerializable` to improve cache key creation

### LINKED ISSUE #669: Paginators Cache ID is too generic
Hey there,

it seems like the ` _getCacheInternalId()` Method in `laminas-paginator/src/Paginator.php` does create too generic cache ids. I use `DoctrinePaginator` with multiple `WHERE` parts. The generated cache key is the same as the `DoctrinePaginator` without `WHERE` part.

There is an exception for `DbSelect` instances. See https://github.com/laminas/laminas-paginator/pull/6

Is there a way to get around this problem?

FabianKoestring:
fixes #669

Adds `JsonSerializable` to `src/DoctrineORMModule/Paginator/Adapter/DoctrinePaginator.php` to optimize cache-key generation in `laminas/laminas-paginator` `_getCacheInternalId()`.

### DIFF
--- a/src/DoctrineORMModule/Paginator/Adapter/DoctrinePaginator.php
+++ b/src/DoctrineORMModule/Paginator/Adapter/DoctrinePaginator.php
@@ -5,12 +5,13 @@ declare(strict_types=1);
 namespace DoctrineORMModule\Paginator\Adapter;
 
 use Doctrine\ORM\Tools\Pagination\Paginator;
+use JsonSerializable;
 use Laminas\Paginator\Adapter\AdapterInterface;
 
 /**
  * Paginator adapter for the Laminas\Paginator component
  */
-class DoctrinePaginator implements AdapterInterface
+class DoctrinePaginator implements AdapterInterface, JsonSerializable
 {
     /** @var Paginator */
     protected $paginator;
@@ -54,5 +55,16 @@ class DoctrinePaginator implements AdapterInterface
     public function count()
     {
         return $this->paginator->count();
+    }
+
+    /**
+     * @return array{select: string, count_select: int}
+     */
+    public function jsonSerialize(): array
+    {
+        return [
+            'select' => $this->paginator->getQuery()->getSQL(),
+            'count_select' => $this->paginator->count(),
+        ];
     }
 }
--- a/test/DoctrineORMModuleTest/Assets/Entity/FormEntity.php
+++ b/test/DoctrineORMModuleTest/Assets/Entity/FormEntity.php
@@ -20,7 +20,7 @@ class FormEntity
     protected $id;
 
     /**
-     * @ORM\Column(type="bool")
+     * @ORM\Column(type="boolean")
      *
      * @var bool
      */
--- a/test/DoctrineORMModuleTest/Assets/Entity/Issue237.php
+++ b/test/DoctrineORMModuleTest/Assets/Entity/Issue237.php
@@ -6,7 +6,7 @@ use Doctrine\ORM\Mapping as ORM;
 
 /**
  * @ORM\Entity
- * @ORM\Table(name="doctrine_orm_module_form_entity")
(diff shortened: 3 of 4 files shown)


driehle:
There are some issues regarding coding standards, could you please have a look at this?

Besides, we should at least have a unit test for the return value of `jsonSerialize()`. This can be done by mocking a paginator.

FabianKoestring:
> There are some issues regarding coding standards, could you please have a look at this?
> 
> Besides, we should at least have a unit test for the return value of `jsonSerialize()`. This can be done by mocking a paginator.

@driehle - I can add tests but why are the `AdapterTest` actually ignored.

driehle:
@FabianKoestring Hm, I have no idea, but I was able to identify the commit that renamed `AdapterTest.php` to `AdapterTestIgnore.php`, which is [db8532d250cec34a9aea4e6ed612baca8e3812e3](https://github.com/doctrine/DoctrineORMModule/commit/db8532d250cec34a9aea4e6ed612baca8e3812e3#diff-8d37c3200d3f5723a8d392df5aedace0e3f2b42ac4a2134b0e441aff1313cd31) from 2012. So it looks like this test has been disabled for quite a long time.

Could you please try enabling that test and see if you can get it running?

Regarding the CS issue, I think PHPCS wants some more information about the type of array. You could use something generic like `@return array<string, mixed>`, but in this case it would be better to use `@return array{select: string, count_select: int}`, as will provide static code analysis much more information. You can run `composer cs-check` locally to see if all issues were fixed.

driehle:
Ok, here are the changes required to make this test run again:

- in `test/DoctrineORMModule/Assets/Entity/FormEntity.php` line 23 `bool` has to be changed to `boolean`
- in `test/DoctrineORMModule/Assets/Entity/Issue237.php` line 9 `doctrine_orm_module_form_entity` has to be changed to `doctrine_orm_module_form_entity_issue237`
- class `DoctrineORMModuleTest\Paginator\AdapterTestIgnore` should be renamed to `DoctrineORMModuleTest\Paginator\AdapterTest`

This gets it running for me. Try running `composer test` locally.

FabianKoestring:
@driehle There are still some problems with php7.4.

driehle:
@FabianKoestring You can see the output [here](https://github.com/doctrine/DoctrineORMModule/pull/670/checks?check_run_id=3924377848), the failing job is `PHPUnit (7.4, highest, false)`. The parameters in brackets come from the [build matrix](https://github.com/doctrine/DoctrineORMModule/blob/4.0.x/.github/workflows/continuous-integration.yml) and stand for `php-version=7.4`, `dependencies=highest` and `option_dependencies=false`. The output is:

```
1) DoctrineORMModuleTest\Paginator\AdapterTest::testCanSetPaginator
Error: Class 'Doctrine\Common\DataFixtures\Loader' not found

/home/runner/work/DoctrineORMModule/DoctrineORMModule/test/DoctrineORMModuleTest/Paginator/AdapterTest.php:33
```

(repeated five times)

This is because in that job optional dependencies (`doctrine/migrations` and `doctrine/data-fixtures`) are removed prior to testing. I suggest you have a look [here](https://github.com/doctrine/DoctrineORMModule/blob/4.0.x/test/DoctrineORMModuleTest/Service/MigrationsCommandFactoryTest.php#L50) how to use `$this->markTestAsIncomplete()` when `doctrine/data-fixtures` is not available. You can simply check for the existence of `Doctrine\Common\DataFixtures\Loader`. You will likely need to do this as early as in `setUp()`.

FabianKoestring:
@driehle - Now everything should fit. Thank you for your support.

driehle:
Indeed, now it's good! Your commits should be squashed though, it I'll take care of that prior to merging.

Thank you for your time and effort, @FabianKoestring!

driehle (approved):

TomHAnderson (approved):

PR #675, opened 2021-10-27
Added support for DBAL v3

snapshotpl:


SenseException:
Please adapt your commit messages as in the description of https://www.doctrine-project.org/contribute/index.html#working-on-topic-branches. Some of the commits can be stashed into one commit, like the `Empty password`for example.

driehle:
We will need to run unit tests with both dbal 2 and 3, as done in https://github.com/doctrine/DoctrineModule/blob/4.2.x/.github/workflows/continuous-integration.yml with laminas form. But first, commits should be squashed as suggested by SenseException. Please also squash the Cs fixes into a single commit.

SenseException:
> But first, commits should be squashed as suggested by SenseException.

Of course squashed, not stashed like I wrote. 😳

snapshotpl:
@driehle @SenseException All tests pass! Looks like required jobs needs to be reconfigure out of repository

.github/workflows/continuous-integration.yml
```diff
@@ -23,14 +22,15 @@ jobs:
           - "7.3"
           - "7.4"
           - "8.0"
+        dbal-version:
+          - "^2.12"
+          - "^3.0"
         dependencies:
           - "highest"
+          - "lowest"
```
driehle:
```suggestion
```

At least in `DoctrineModule`, we do only test lowest dependencies for the lowest PHP version, but not for all PHP versions. I think this would make sense here as well, as otherweise we have about 30 checks to run.

.github/workflows/continuous-integration.yml
```diff
@@ -23,14 +22,15 @@ jobs:
...
         optional_dependencies:
           - true
           - false
-        include:
-          - php-version: "7.3"
-            dependencies: "lowest"
```
driehle, line 50:
```suggestion
        include:
          - php-version: "7.3"
            dependencies: "lowest"
            dbal-version: "^2.12"
            optional_dependencies: false
          - php-version: "7.3"
            dependencies: "lowest"
            dbal-version: "^3.0"
            optional_dependencies: false
          - php-version: "7.3"
            dependencies: "lowest"
            dbal-version: "^2.12"
            optional_dependencies: true
          - php-version: "7.3"
            dependencies: "lowest"
            dbal-version: "^3.0"
            optional_dependencies: true
            
```

If we don't want to run lowest dependencies for all PHP version, we need to add these checks explicitly, as above.

config/module.config.php
```diff
@@ -292,3 +289,9 @@
         ],
     ],
 ];
+
+if (class_exists(ImportCommand::class)) {
+    $result['service_manager']['invokables']['doctrine.dbal_cmd.import'] = ImportCommand::class;
+}
+
+return $result;
```
driehle, line 298: Why are we not promoting `Doctrine\DBAL\Tools\Console\Command\ReservedWordsCommand` here as well? <br>Available with [2.13](https://github.com/doctrine/dbal/blob/2.13.x/lib/Doctrine/DBAL/Tools/Console/Command/ReservedWordsCommand.php) and [3.1](https://github.com/doctrine/dbal/blob/3.1.x/src/Tools/Console/Command/ReservedWordsCommand.php)

src/DoctrineORMModule/Options/DBALConnection.php
```diff
@@ -4,10 +4,13 @@
 
 namespace DoctrineORMModule\Options;
 
-use Doctrine\DBAL\Driver\PDOMySql\Driver;
+use Doctrine\DBAL\Driver\PDO\MySQL\Driver as DriverV3;
+use Doctrine\DBAL\Driver\PDOMySql\Driver as DriverV2;
```
driehle: I don't think that this is needed here. The class `Doctrine\DBAL\Driver\PDO\MySQL\Driver` is already available in DBAL 2 as well, see for https://github.com/doctrine/dbal/blob/2.13.x/lib/Doctrine/DBAL/Driver/PDO/MySQL/Driver.php. I think it has been added in 2.11 according to [this commit](https://github.com/doctrine/dbal/commit/67d6bb0395d3e3e1aadfd23f5658a2a846266dfd). So we should probably be save only using `PDO\MySQL`?

test/testing.config.php
```diff
@@ -1,6 +1,7 @@
 <?php
 
-use Doctrine\DBAL\Driver\PDOSqlite\Driver;
+use Doctrine\DBAL\Driver\PDO\SQLite\Driver;
+use Doctrine\DBAL\Driver\PDOSqlite\Driver as DriverV2;
```
driehle: Again, I think we can use the new name only here.

driehle (changes requested):

composer.json
```diff
@@ -50,17 +50,17 @@
     "require":           {
         "php":                                  "^7.3 || ^8.0",
         "composer/package-versions-deprecated": "^1.11",
-        "doctrine/dbal":                        "^2.10",
+        "doctrine/dbal":                        "^2.12 || ^3.0",
```
driehle: Is there a certain reason for 2.12 and 3.0? These versions seem to be unsupported, so maybe we should use ^2.13 and ^3.1?

.github/workflows/continuous-integration.yml
snapshotpl: Ok, I can do it, but I would like to understand what is wrong with 30 checks? It's add little more time to wait but also proves that works in many strange version combination. And it works now.

src/DoctrineORMModule/Options/DBALConnection.php
snapshotpl: I had problems with v2 and v3 compatibility, but in the meantime I up minimum version, so maybe now it works. To check

composer.json
snapshotpl: Doctrine documentation says 2.10.2 is current version (probably wrong). Anyway, if we can keep older version in low cost (like now), that we can do it, because why not?

driehle: Could it be that you accidentaly looked at orm and not a dbal? For orm 2.10.2 is indeed current, but for dbal https://www.doctrine-project.org/projects/dbal.html says that 2.13.4 and 3.1.3 are both current and supported.

driehle:
@snapshotpl regarding https://github.com/doctrine/DoctrineORMModule/pull/675#discussion_r742035456 and https://github.com/doctrine/DoctrineORMModule/pull/675#discussion_r742039430:

dbal 2.x includes a forward compatibility layer to dbal 3.x. That means, the new symbols from 3.x have been backported to 2.x so that 2.x includes the old symbols (which are then deprecated) and the new symbols. Users should, therefore, first go the the latest 2.x release, have their code being checked for deprecated symbols (using static analysis), fix their code to not use any deprecated symbols anymore and, then, can hopefully upgrade to 3.x without any BC breaks.

A lot of the deprecations came in [dbal 2.11.0](https://github.com/doctrine/dbal/releases/tag/2.11.0), but some more in [dbal 2.13.0](https://github.com/doctrine/dbal/releases/tag/2.13.0), see especially [this PR](https://github.com/doctrine/dbal/pull/4529). Therefore, I think we should bumb dbal to ^2.13.0. We can keep ^3.0.0 though I don't really like to release new support for already unmaintained versions and, hence, would recommend ^3.1.0 there as well.

src/DoctrineORMModule/Options/DBALConnection.php
driehle: As pointed out in https://github.com/doctrine/DoctrineORMModule/pull/675#issuecomment-960528042, raising the minimum dbal version brings in a forward compatibility layer in dbal 2.x and, hence, allows using the new symbols of 3.x in 2.x already.

.github/workflows/continuous-integration.yml
driehle: @greg0ire I think we discussed this earlier and you wanted me to test lowest dependencies only with the lowest supported PHP version. What do you think of this, should we test lowest dependencies with all supported PHP versions? As we support both dbal 2 and 3, we have all tests running for both dbal 2 and 3. This results in 30 checks alltogether.

.github/workflows/continuous-integration.yml
```diff
@@ -23,14 +22,15 @@ jobs:
           - "7.3"
           - "7.4"
           - "8.0"
+        dbal-version:
+          - "^2.13"
+          - "^3.1"
```
driehle, line 27:
```suggestion
          - "2.13.0"
          - "3.1.0"
```

.github/workflows/continuous-integration.yml
```diff
@@ -58,6 +58,9 @@ jobs:
           ini-values: "zend.assertions=1"
           extensions: "pdo_mysql"
 
+      - name: "Install desired DBAL version"
+        run: "composer require doctrine/dbal ${{ matrix.dbal-version }}"
```
driehle:
```suggestion
      - name: "Require specific DBAL version"
        run: "composer require doctrine/dbal ^${{ matrix.dbal-version }} --no-update"
```

.gitignore
```diff
@@ -8,3 +8,5 @@ composer.phar
 .phpunit.result.cache
 .php_cs.cache
 .phpcs-cache
+config/application.config.php
+ci/20*/
```
driehle, line 12: What is this line for?

driehle (changes requested):
Two remarks to bring the workflow in line with https://github.com/doctrine/DoctrineModule/blob/4.2.x/.github/workflows/continuous-integration.yml.

.gitignore
snapshotpl: Running test local generate migration like `ci/2020/Version20211102200030.php` (current date)

.github/workflows/continuous-integration.yml
snapshotpl: Not sure here. It's locks hard tested versions. `lowest` already check `3.1.0` and ``highest` test any future version.

.github/workflows/continuous-integration.yml
greg0ire: Yes, I think one `lowest` build should be enough to figure out if you need to bump version constraints: if you need a specific commit, then you can figure out since which version it's available on all major versions. There are infinite version combinations, let's only test the cases that will catch the most bugs.

snapshotpl: Now CI config file is simple and any change in matrix will be correctly respect.
Removed version of `include` contains one position but missing `optional_dependencies`, because someone miss it. Too much work for us to maintain this IMO in future.

However I can change this if you still think to remove this.

snapshotpl:
@SenseException @driehle @greg0ire I need final decision what we do with lowest/latest and I'd like to merge to prevent more conflicts ;)

driehle:
@snapshotpl sorry for the waiting. I wanted to bring in some other PRs first, to prepare DoctrineORMModule for the next minor release, before asking you to further work on this. These have all been merged now. In order to bring this PR forward, please:

- rebase on current `4.1.x`
- have `lowest` dependencies only running for PHP 7.3
- squash your commits, especially these small fixes should go into a respective previous commit

driehle:
The rebase looks a bit flawed, the first 7 commits shouldn't be there... 🤔

snapshotpl:
Yeah, no idea what happen. I don't use squash in my daily work.

snapshotpl:
@driehle fixed!

driehle:
I just rebased your changes to the latest `4.1.x`. LGTM! <br>
I'll ask for a second review though.

driehle (approved):

src/Service/RunSqlCommandFactory.php
```diff
@@ -25,6 +25,7 @@ public function __invoke(ContainerInterface $serviceLocator, $requestedName, ?ar
             );
         }
 
+        /** @psalm-suppress TooFewArguments */
```
SenseException: Instead of Psalm annotations this should be put into the psalm config.

SenseException (changes requested):

src/Service/RunSqlCommandFactory.php
driehle: Do you mean putting it in the baseline?

SenseException: Yes, looks like the baseline file is supposed to contain it.

driehle: Actually, I removed the whole check for the presence of SingleConnectionProvider, as that was introduced in dbal 2.11.0 (see [dbal@b10d4db8](https://github.com/doctrine/dbal/commit/b10d4db882eeaeda044aa71946d87882a244a094)) and we do not support earlier versions anymore.

SenseException (approved):

PR #677, opened 2021-11-05
Removed UPGRADE.md, see Github releases instead

driehle:


greg0ire (approved):
:thinking: note that on other repositories, we use that file to document BC-Breaks (so it's basically an upgrade guide)

Don't you have that need on this repository?

driehle:
Pre 4.0 releases are still from a Zend Framework era, so currently there is indeed no need for this. Furthermore, BC Breaks are tagged with "BC Break" and, hence, appear in the release notes as well in their own section.

PR #679, opened 2021-11-05
Updated docs and readme to include installation notes for PHP 8

driehle:
Since end users might not directly look into the documentation of `DoctrineModule`, I have added the installation notes regarding PHP 8 to the documentation of `DoctrineORMModule` as well.

Technically, the troubles installing with PHP 8 come from the laminas-cache integration in `DoctrineModule`, but since that is a requirement of `DoctrineORMModule`, the same limitations apply here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,14 @@
 Doctrine ORM Module for Laminas
 ===============================
 
-[![Build Status](https://github.com/doctrine/DoctrineORMModule/workflows/Continuous%20Integration/badge.svg)](https://github.com/doctrine/DoctrineORMModule/actions)
-[![Code Coverage](https://codecov.io/gh/doctrine/DoctrineORMModule/branch/4.0.x/graph/badge.svg)](https://codecov.io/gh/doctrine/DoctrineORMModule/branch/4.0.x)
+[![Build Status](https://github.com/doctrine/DoctrineORMModule/actions/workflows/continuous-integration.yml/badge.svg)](https://github.com/doctrine/DoctrineORMModule/actions/workflows/continuous-integration.yml?query=branch%3A4.1.x)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -21,8 +23,40 @@ The sources files for the documentation can be found in the [`docs` dir](./docs/
...
 ```
+
+### Note on PHP 8.0 or later
+
+[DoctrineModule](https://github.com/doctrine/DoctrineModule/) provides an integration with 
+[laminas-cache](https://docs.laminas.dev/laminas-cache/), which currently comes with some storage adapters which 
+are not compatible with PHP 8.0 or later. To prevent installation of these unused cache adapters, you will need 
+to add the following to your `composer.json` file:
+
+```
```
greg0ire:
```suggestion
```json
```

docs/en/index.rst
```diff
@@ -1,8 +1,55 @@
...
+Note on PHP 8.0 or later
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+`DoctrineModule <https://www.doctrine-project.org/projects/doctrine-module/en/current/index.html>`__
+provides an integration with `laminas-cache <https://docs.laminas.dev/laminas-cache/>`__, which
+currently comes with some storage adapters which are not compatible with PHP 8.0 or later. To
+prevent installation of these unused cache adapters, you will need to add the following to your
+``composer.json`` file:
+
+.. code::
```
greg0ire: How can we let RST know this is JSON

driehle: I'd assume that it is `.. code:: json`, but I am not sure.

greg0ire (approved):
Please don't resolve conversations you don't start :pray:

PR #680, opened 2021-11-05
Remove PHP version config from `composer.json`

driehle:
The PHP version should not be set to a simulated version, but rather be the real version tests are running with. This requries --ignore-platform-req=php for tests with PHP 8.0 or later. See installation notes of DoctrineModule or laminas-cache, which explains why this is needed to get legacy cache adapters running.

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -1,41 +1,16 @@
-
 name: "Coding Standards"
 
 on:
   pull_request:
     branches:
       - "*.x"
-      - "master"
   push:
     branches:
       - "*.x"
-      - "master"
 
 jobs:
   coding-standards:
     name: "Coding Standards"
-    runs-on: "ubuntu-20.04"
-
-    strategy:
-      matrix:
-        php-version:
-          - "7.4"
-          - "8.0"
-
-    steps:
-      - name: "Checkout"
-        uses: "actions/checkout@v2"
-
-      - name: "Install PHP"
-        uses: "shivammathur/setup-php@v2"
-        with:
-          coverage: "none"
-          php-version: "${{ matrix.php-version }}"
-          tools: "cs2pr"
-
-      - name: "Install dependencies with Composer"
-        uses: "ramsey/composer-install@v1"
-
-      # https://github.com/doctrine/.github/issues/3
-      - name: "Run PHP_CodeSniffer"
-        run: "vendor/bin/phpcs -q --no-colors --report=checkstyle | cs2pr"
+    uses: "doctrine/.github/.github/workflows/coding-standards.yml@1.2.0"
+    with:
+      php-version: '7.4'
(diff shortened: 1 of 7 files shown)


.github/workflows/coding-standards.yml
```diff
@@ -1,41 +1,16 @@
...
-          php-version: "${{ matrix.php-version }}"
-          tools: "cs2pr"
-
-      - name: "Install dependencies with Composer"
-        uses: "ramsey/composer-install@v1"
-
-      # https://github.com/doctrine/.github/issues/3
-      - name: "Run PHP_CodeSniffer"
-        run: "vendor/bin/phpcs -q --no-colors --report=checkstyle | cs2pr"
+    uses: "doctrine/.github/.github/workflows/coding-standards.yml@1.1.1"
```
greg0ire: Let's use the latest version

```suggestion
    uses: "doctrine/.github/.github/workflows/coding-standards.yml@1.2.0"
```

.github/workflows/coding-standards.yml
```diff
@@ -1,41 +1,16 @@
...
+    with:
+      php-version: '7.4'
```
greg0ire, line 16: You can omit these 2 lines, and it will run with 8.0 (and it will probably pass)

.github/workflows/release-on-milestone-closed.yml
```diff
@@ -8,38 +8,10 @@ on:
...
-      - name: "Create new milestones"
-        uses: "laminas/automatic-releases@v1"
-        with:
-          command-name: "laminas:automatic-releases:create-milestones"
-        env:
-          "GITHUB_TOKEN": ${{ secrets.GITHUB_TOKEN }}
-          "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
-          "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
-          "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
+    uses: "doctrine/.github/.github/workflows/release-on-milestone-closed.yml@1.1.1"
```
greg0ire:
```suggestion
    uses: "doctrine/.github/.github/workflows/release-on-milestone-closed.yml@1.2.0"
```

.github/workflows/coding-standards.yml
driehle: I tried that, but it won't pass with 8.0, because of the whole setup not being installable under 8.0 without `--ingnore-platform-arg=php`. For the sake of simplicity, I am keeping coding standards at 7.4 currently. Both here and upstream in `DoctrineModule`.

greg0ire: Ok great 👍

greg0ire (approved):

PR #683, opened 2021-11-08
Update dependencies and allow PHP 8.1

### LINKED ISSUE #676: Support for PHP 8.1
Since doctrine-laminas-hydrator and DoctrineModule in their latest version are now PHP 8.1-compatible, we should add this in DoctrineORMModule as well.

Similar to DoctrineModule that will not yet allow end users to upgrade to PHP 8.1 due to missing support in upstream dependencies, but we can set up unit tests already.

driehle:
This PR updates the dependencies of DoctrineORMModule by raising the versions of the requirements:

- DoctrineModule is raised from ^4.0 to ^4.2.1 for PHP 8.1 support and latest version of doctrine-laminas-hydrator supporting embeddables
- doctrine/orm is raised from ^2.6.4 to ^2.9.6, mostly because 2.9.0 is buggy in regard to the ClearMetadataCache command
- doctrine/persistence is raised from ^1.3.3||^2.0 to ^2.2.2, because doctrine/orm >= 2.9 requires persistence 2.x anyways, so persistence 1.x is not installable anymore
- composer/package-versions-deprecated is dropped, because it's not used anymore

With these updated dependencies, DoctrineORMModule runs under PHP 8.1. Unit tests for PHP 8.1 are added and compatibility with PHP 8.1 is declared. Note that some upstream libraries are not yet PHP 8.1-compatible, so end users will not yet be able to install DoctrineORMModule with PHP 8.1. This is the same as in DoctrineModule, see https://github.com/doctrine/DoctrineModule/pull/741.


TomHAnderson (approved):

driehle:
I found that laminas/laminas-code, laminas/laminas-console and laminas/laminas-mvc-console can be removed from require-dev as well, as they are not needed in the unit tests. I have added another commit for this.

SenseException (approved):
When phpstan gets introduced, the .github static analysis workflow could be introduced.

driehle:
@SenseException 

> When phpstan gets introduced, the .github static analysis workflow could be introduced.

I'll introduce PhpStan in one of the next PRs. However, there is an issue with the upstream workflow: It doesn't allow to specify composer options and Psalms minimum PHP version via inputs. See https://github.com/laminas/laminas-code/issues/67 for further details. The issue is that some dependencies start using PHP 7.4-only code in their latest version, but Psalm inflects PHP 7.3 from our `composer.json`. So we need to run Psalm with `--php-version=8.0`, otherwise there will by parse errors.

@greg0ire Do you think we could add two input variables `psalm-php-version` and `composer-options` to the `static-analysis.yml` template in the upstream workflows?

greg0ire:
@driehle maybe we should use this approach instead: https://github.com/doctrine/mongodb-odm/pull/2268 ? I don't see why you would want to specify another version than the one you are currently using :thinking:

driehle:
@greg0ire Yes, I think that would work. Still we'd need `composer-options` here, because the whole setup is not installable because of the [laminas-cache issue](https://www.doctrine-project.org/projects/doctrine-module/en/4.2/index.html#note-on-php-8-0-or-later). Or we'd need to inject a `replace` section into `composer.json` in CI, though I don't know how to do this.

greg0ire:
`composer-options` sounds good, yeah 👍

driehle:
This is something that could help us get rid of `--ignore-platform-req=php`:

```bash
cat composer.json \
  | jq '. * {replace: {"laminas/laminas-cache-storage-adapter-apc": "*"}}' \
  | jq '. * {replace: {"laminas/laminas-cache-storage-adapter-dba": "*"}}' \
  | jq '. * {replace: {"laminas/laminas-cache-storage-adapter-memcache": "*"}}' \
  | jq '. * {replace: {"laminas/laminas-cache-storage-adapter-memcached": "*"}}' \
  | jq '. * {replace: {"laminas/laminas-cache-storage-adapter-mongodb": "*"}}' \
  | jq '. * {replace: {"laminas/laminas-cache-storage-adapter-wincache": "*"}}' \
  | jq '. * {replace: {"laminas/laminas-cache-storage-adapter-xcache": "*"}}' \
  | jq '. * {replace: {"laminas/laminas-cache-storage-adapter-zend-server": "*"}}'
```

I'll try that in a different PR though.

**Edit:** Nah, though that would be much nicer than `--ignore-platform-req=php`, it still would only work if there was a way to inject this into the upstream workflows prior to running `composer install`.

PR #684, opened 2021-11-08
inject replace section regarding laminas-cache in CI

driehle:


driehle:
@greg0ire @SenseException Just for the records, I tried it, and manipulating `composer.json` in CI works:

```yml
      - name: "Inject replace section into composer.json"
        run: |
          cat <<< $(jq '. * {replace: {
            "laminas/laminas-cache-storage-adapter-apc": "*",
            "laminas/laminas-cache-storage-adapter-dba": "*",
            "laminas/laminas-cache-storage-adapter-memcache": "*",
            "laminas/laminas-cache-storage-adapter-memcached": "*",
            "laminas/laminas-cache-storage-adapter-mongodb": "*",
            "laminas/laminas-cache-storage-adapter-wincache": "*",
            "laminas/laminas-cache-storage-adapter-xcache": "*",
            "laminas/laminas-cache-storage-adapter-zend-server": "*"
          }}' composer.json) > composer.json
          cat composer.json
```

However, in the current case this doesn't make things simpler. Since we still need `--ignore-platform-req=php` for PHP 8.1, I do not gain anything by removing it for PHP 8.0 using the `replace` section.

PR #685, opened 2021-11-09
Updated doctrine/coding-standard to 9.0

driehle:


snapshotpl (approved):

tests/Assets/Entity/Target.php
```diff
@@ -1,7 +1,9 @@
 <?php
 
+declare(strict_types=1);
+
 namespace DoctrineORMModuleTest\Assets\Entity;
 
-interface TargetInterface
+interface Target
```
SenseException, line 7: Is this BC break not something for a major release?

driehle: I don't think so, as this is in the unit tests and not in the library itself.

driehle: Btw, [semver](https://semver.org/) refers to the "public API" of a project and I don't think unit tests are considered part of the public API. Hence, we should be save changing in the unit tests whatever we want.

SenseException: > this is in the unit tests and not in the library itself.

Ah, I missed that it's a test interface. Sorry for the holdup.

SenseException (approved):

tests/Assets/Entity/Target.php
driehle: No worries 😊

PR #687, opened 2021-11-11
Deprecate `createService()` methods as laminas-servicemanager 3.x is already required

driehle:
This deprecates the `createService()` methods which are a left-over from laminas-servicemanager 2.x. Since we already require laminas-servicemanager 3.x for a while, we should remove these methods in the next major release, hence, in the next minor release we mark them as deprecated. That is what this PR does.

### DIFF
--- a/src/Service/CliConfiguratorFactory.php
+++ b/src/Service/CliConfiguratorFactory.php
@@ -21,6 +21,8 @@ class CliConfiguratorFactory implements FactoryInterface
 
     /**
      * {@inheritDoc}
+     *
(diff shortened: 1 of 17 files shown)


greg0ire (approved):
Looks fine but bear in mind I have no knowledge about Laminas :P

snapshotpl (approved):
As member of laminas I can confirm that's fine

driehle:
We did the same for DoctrineModule already (see https://github.com/doctrine/DoctrineModule/commit/fa2799efb69d89141d9f4cf65c8a4b9b037e1bf2). <br>Thanks for the confirmation, @snapshotpl!

PR #690, opened 2021-11-11
Make static analysis more strict, Psalm level 5, PhpStan level 5

driehle:
This PR increases PhpStan from level 3 to 5 and lowers Psalm from level 7 to 5.

Note that this sets `treatPhpDocTypesAsCertain: false`, as suggested in https://github.com/phpstan/phpstan/issues/1926 for libraries.

### DIFF
--- a/tests/Assets/Entity/EntityWithoutRepository.php
+++ b/tests/Assets/Entity/EntityWithoutRepository.php
@@ -16,7 +16,17 @@ class EntityWithoutRepository
      * @ORM\Column(type="integer")
      * @ORM\GeneratedValue(strategy="AUTO")
      *
-     * @var int
+     * @var ?int
      */
     private $id;
+
+    public function getId()
+    {
+        return $this->id;
+    }
+
+    public function setId($id): void
+    {
+        $this->id = $id;
+    }
 }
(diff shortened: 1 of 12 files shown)


tests/Assets/Entity/EntityWithoutRepository.php
```diff
@@ -16,7 +16,17 @@ class EntityWithoutRepository
      * @ORM\Column(type="integer")
      * @ORM\GeneratedValue(strategy="AUTO")
      *
-     * @var int
+     * @var ?int
      */
     private $id;
+
+    public function getId(): int
```
snapshotpl: If property is nullable so getter type should also.

tests/CliConfiguratorTest.php
```diff
@@ -121,7 +121,7 @@ public function testValidHelpers(): void
 
         $dbHelper = $helperSet->get('db');
         assert($dbHelper instanceof ConnectionHelper);
-        $this->assertInstanceOf(ConnectionHelper::class, $dbHelper);
+        $this->assertInstanceOf(ConnectionHelper::class, $dbHelper); // @phpstan-ignore-line
```
snapshotpl: Why not use baseline or ignore in phpstan config?

driehle: I don't like putting this in the baseline, since for me the baseline is something that I'll drop and re-create with every major release. That way I can ensure to deal with BC breaking improvements for every major. This issue, however, is something that will stay probably forever, simply because PhpStan doesn't understand it.

I agree that putting it in PhpStan config would be better, I'll do that.

TomHAnderson (approved):

PR #698, opened 2021-12-15
Fixed deprecation on Serializable interface

### LINKED ISSUE #695: Deprecation of Serializable interface in PHP 8.1
`Deprecated: DoctrineORMModule\Collector\MappingCollector implements the Serializable interface, which is deprecated. Implement __serialize() and __unserialize() instead (or in addition, if support for old PHP versions is necessary) in /code/vendor/doctrine/doctrine-orm-module/src/Collector/MappingCollector.php on line 21`

driehle:
fixes #695 

This sets `convertDeprecationsToExceptions="true"` in PHPUnit configuration, so that in the future we can see failings tests when there are deprecation messages in PHP.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -70,7 +70,7 @@ jobs:
         with:
           php-version: "${{ matrix.php-version }}"
           coverage: "pcov"
-          ini-values: "zend.assertions=1"
+          ini-values: "error_reporting=E_ALL, zend.assertions=1"
           extensions: "pdo_mysql"
 
       - name: "Require specific DBAL version"
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -7,6 +7,7 @@
     convertErrorsToExceptions="true"
     convertNoticesToExceptions="true"
     convertWarningsToExceptions="true"
+    convertDeprecationsToExceptions="true"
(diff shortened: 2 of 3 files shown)


driehle:
@greg0ire While working on #695 I found that Github actions won't fail unit tests when deprecation messages occur (see https://github.com/doctrine/DoctrineORMModule/runs/4538099900?check_suite_focus=true). Is there a certain reason for this, i.e. is this the intended behaviour?

I think it would make sense to change this. I thought setting `convertDeprecationsToExceptions="true"` in `phpunit.xml.dist` would be enough, but it looks like `error_reporting=E_ALL` has to be set as well. With this, tests are now failing (https://github.com/doctrine/DoctrineORMModule/runs/4539273011?check_suite_focus=true) as I would have expected it.

greg0ire:
I'd say yes, it's intended: the build should not suddenly fail from one day to the other. I'd say you can change that if you feel up to dealing with deprecations as they occur. Also, I know that in some repositories, we do fail the build if static analysis detects deprecations (which we can ignore or fix on a case by case basis).

It might be a better experience to have a non-required job fail if a previous phpunit job had deprecations, that way you are aware of the issue without it blocking the merge.

TomHAnderson (approved):

SenseException (approved):

PR #704, opened 2022-01-03
Removed signatures related to laminas-servicemanager 2.x 

driehle:
In laminas-servicemanager 2.x, the `FactoryInterface` included the following methods:
 - `public function createService(ServiceLocatorInterface $serviceLocator)`
 - `public function setCreationOptions(array $options)`

These methods are replaced by the following method in laminas-servicemanager 3.x:
 - `public function __invoke(ContainerInterface $container, $requestedName, ?array $options = null)`

Since DoctrineORMModule already requires laminas-servicemanager 3.x since version 3.0.0, i.e. for nearly two years, it is time to drop the obsolete functions `createService()` and `setCreationOptions()` in all factories. All factories already implement a `__invoke()` method.

This is the same as done in doctrine/DoctrineModule#763

### DIFF
--- a/tests/Service/ConfigurationFactoryTest.php
+++ b/tests/Service/ConfigurationFactoryTest.php
@@ -12,6 +12,7 @@ use Doctrine\ORM\Mapping\EntityListenerResolver;
 use Doctrine\ORM\Mapping\NamingStrategy;
 use Doctrine\ORM\Mapping\QuoteStrategy;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
+use DoctrineORMModule\Options\Configuration;
 use DoctrineORMModule\Service\ConfigurationFactory;
 use DoctrineORMModuleTest\Assets\RepositoryClass;
 use Laminas\ServiceManager\Exception\InvalidArgumentException;
@@ -47,7 +48,7 @@ class ConfigurationFactoryTest extends TestCase
             ],
         ];
         $this->serviceManager->setService('config', $config);
-        $ormConfig = $this->factory->createService($this->serviceManager);
+        $ormConfig = ($this->factory)($this->serviceManager, Configuration::class);
         $this->assertInstanceOf(NamingStrategy::class, $ormConfig->getNamingStrategy());
     }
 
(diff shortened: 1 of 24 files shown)


tests/Service/ConfigurationFactoryTest.php
```diff
@@ -47,7 +48,7 @@ public function testWillInstantiateConfigWithoutNamingStrategySetting(): void
             ],
         ];
         $this->serviceManager->setService('config', $config);
-        $ormConfig = $this->factory->createService($this->serviceManager);
+        $ormConfig = $this->factory->__invoke($this->serviceManager, Configuration::class);
```
TomHAnderson: Calling __invoke directly is unusual.  I recommend 
```php
$ormConfig = $this->factory($this->serviceManager, Configuration::class);
```

tests/Service/ConfigurationFactoryTest.php
```diff
@@ -64,7 +65,7 @@ public function testWillInstantiateConfigWithNamingStrategyObject(): void
         ];
         $this->serviceManager->setService('config', $config);
         $factory   = new ConfigurationFactory('test_default');
-        $ormConfig = $factory->createService($this->serviceManager);
+        $ormConfig = $factory->__invoke($this->serviceManager, Configuration::class);
```
TomHAnderson: Same comment for all __invoke calls.  Change them to `$factory(params)`

TomHAnderson (changes requested):
The PR is good except for the `->__invoke` calls.

tests/Service/ConfigurationFactoryTest.php
driehle: You cannot use `$this->factory()` here, as then PHP tries to call `ConfigurationFactoryTest::factory()` instead of `ConfigurationFactoryTest::$factory()`. I could use `($this->factory)()` though, which still looks a bit akward to me.

tests/Service/ConfigurationFactoryTest.php
driehle: Fair enough, this one can be rewritten.

tests/Service/ConfigurationFactoryTest.php
TomHAnderson: I think `($this->factory)()` is perfectly acceptable.  I resolve objects this way all the time and that's especially true in Doctrine.

```php
$entity = (new Entity()) 
    ->setName('name')
    ;
```

is a common pattern in my work.  I certainly think `($this->factory)()` is much more readable than directly calling `__invoke`.

SenseException (approved):

tests/Service/ConfigurationFactoryTest.php
driehle: Added with 46b6dc2, requesting approval 😉

TomHAnderson (approved):

PR #709, opened 2022-01-21
Add support for schema assets filter option

villermen:
`$filterSchemaAssetsExpression` was deprecated, but its `$schemaAssetsFilter` callback replacement hasn't been implemented in doctrine-orm-module yet. This PR tries to correct that by defining the option on `Configuration`.

If there is any additional documentation/testing needed just point me in the right direction and I'll get that sorted!

### DIFF
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -299,9 +299,9 @@ The "schema_assets_filter" option can be used to exclude certain tables from bei
         'doctrine' => [
             'configuration' => [
                 'orm_default' => [
-                    'schema_assets_filter' => static function ($tableName) {
-                        return $tableName === 'foobar';
-                    }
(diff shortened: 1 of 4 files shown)


SenseException:
In case of documentation: https://github.com/doctrine/DoctrineORMModule/blob/5.1.x/docs/en/configuration.rst

villermen:
I've added a unit test and documentation example!

SenseException (approved):

driehle:
@villermen Could you please rebase your changes on the latest `5.1.x` branch? This should solve the broken CI pipelines.

villermen:
@driehle Done!

driehle (approved):

driehle:
Thank you for your time and contribution, @villermen!

It will probably take a few days for the release of 5.1.0 though, but you can give 5.1.x-dev a try if you like.

PR #722, opened 2022-08-15
Set version_column_length to 191 to work with MySQL 8.0 and MariaDB 10.5

### LINKED ISSUE #720: "Specified key was too long; max key length is 767 bytes" - for version column
The default configured [version_column_length in module.config.php](https://github.com/doctrine/DoctrineORMModule/blob/5.3.x/config/module.config.php#L157) of 1024 is too long in MySQL/MariaDB with utf8mb4 (which is the default in MySQL 8.0 and MariaDB 10.5 and later)

This gives an error when running migrations:migrate for the first time or migrations:sync-metadata-storage when metadata storage needs to be synchronized

In doctrine/migrations it was reduced to 191.
See https://github.com/doctrine/migrations/issues/958 and https://github.com/doctrine/migrations/pull/960/files

ittmann:
Lower version_column_length from 1024 tot 191
Fixes https://github.com/doctrine/DoctrineORMModule/issues/720

### DIFF
--- a/ci/config/ci.config.php
+++ b/ci/config/ci.config.php
@@ -27,7 +27,7 @@ return [
                 'table_storage' => [
                     'table_name' => 'DoctrineMigrationVersions',
                     'version_column_name' => 'version',
-                    'version_column_length' => 1024,
+                    'version_column_length' => 191,
                     'executed_at_column_name' => 'executedAt',
                     'execution_time_column_name' => 'executionTime',
                 ],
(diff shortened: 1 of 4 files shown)


driehle (changes requested):
Thank you for your PR. The failed CI pipeline seems to be unrelated and should be fixed by #723. Could you please rebase your changes on the current `5.2.x` branch to see if this fixes your CI issue?

Furthermore, besides `config/module.config.php` there are a few other places where the column lenght should be changed:
- `ci/config/ci.config.php`
- `docs/en/migrations.rst`
- `tests/testing.config.php` (two times)

Please include these changes as well.

driehle (approved):
LGTM 👍

driehle:
Closed with 323eccfae756650c7ed8b6996c9c9ff37095ad03

PR #728, opened 2022-12-12
PHP 8.2 Support

rogervila:
Hi,

This PR needs https://github.com/doctrine/DoctrineModule to release a 5.3.x version to support PHP 8.2

PR #729, opened 2022-12-21
Drop `container-interop/container-interop`

tomudding:
As suggested in #728 this drops the abandoned `container-interop/container-interop` in favour of `psr/container`. This means that the `Interop\Container\ContainerInterface` interface is replaced with `Psr\Container\ContainerInterface`.

Unlike doctrine/DoctrineModule#792 not all changed signatures are in `final` classes, specifically the `CliConfigurator`. This may require bumping `laminas/laminas-servicemanager` to at least `v3.11.0`, @driehle please advise.


tomudding:
I am aware of the failing tests, I will address these in a bit.

fezfez:
seem good to me :1st_place_medal:

driehle:
> This may require bumping `laminas/laminas-servicemanager` to at least `v3.11.0`, @driehle please advise.

Indeed, we need to bump laminas-servicemanager to the version introducing class aliases for the interop and the Psr containers. For the records, boesing has explained this here: https://github.com/php-fig/container/issues/37

Please bump to ^3.17.0. Otherwise this PR looks fine.

tomudding:
> Please bump to ^3.17.0. Otherwise this PR looks fine.

I have amended my first commit to bump the version to ensure correct history. Thank you for your time taking a look at this :)

driehle (approved):

PR #731, opened 2023-01-25
Fix schema_assets_filter configuration docs

ittmann:
The callable set as schemaAssetsFilter should return false for tables that you want filtered out 

See for example https://github.com/doctrine/dbal/blob/3.6.x/src/Configuration.php#L168 and https://github.com/doctrine/dbal/blob/4.0.x/src/Schema/AbstractSchemaManager.php#L204

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -15,9 +15,5 @@ parameters:
         -
             message: '#Parameter \#1 .* of method DoctrineORMModule\\Options\\Configuration.* stdClass given#'
             path: tests/Options/ConfigurationOptionsTest.php
-        -
-            message: '#Parameter \#1 .* of method Doctrine\\Common\\DataFixtures\\Executor\\ORMExecutor::execute\(\)#'
-            path: tests/Paginator/AdapterTest.php
 includes:
     - vendor/phpstan/phpstan-phpunit/extension.neon
-
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -293,7 +293,8 @@ module.
 How to Exclude Tables from a Schema Diff
 ----------------------------------------
 
-The "schema_assets_filter" option can be used to exclude certain tables from being created or deleted in a schema update:
+The "schema_assets_filter" option can be used to exclude certain tables from being deleted in a schema update.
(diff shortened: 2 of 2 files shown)


SenseException:
A description about what to expect on true and false would be insightful. Something like "If the function returns true, then..."

driehle (changes requested):
Yes, please add something in the writting text as well on how this filter works!

DIReports:
My intent with this PR was to fix the documentation error.
Documentation should probably be added to the [DBAL Configuration documentation](https://www.doctrine-project.org/projects/doctrine-dbal/en/current/reference/configuration.html) and/or [Configuration class](https://github.com/doctrine/dbal/blob/3.5.x/src/Configuration.php#L44)

With the latest package versions (dbal 3.5.3 and migrations 3.5.5) adding the 'migrations' table to the asset filter breaks doctrine migrations (and not adding it makes the schema-tool when run with the the "--complete" option drop the table)

DIReports:
Adding a table to the schema_asset_filter will not prevent an entity table from being created, only tables not owned by the DBAL from being dropped, I changed this in the first line.

Also I added some explanation on how the schema_assets_filter works and removed the 'migrations' table from the example as adding it to the asset filter currently breaks doctrine migrations.

ittmann:
> With the latest package versions (dbal 3.5.3 and migrations 3.5.5) adding the 'migrations' table to the asset filter breaks doctrine migrations (and not adding it makes the schema-tool when run with the the "--complete" option drop the table)

Opened issue https://github.com/doctrine/migrations/issues/1317
**Adding the migrations table to doctrine (dbal) schema assets filter breaks migrations**

SenseException (approved):

TomHAnderson (approved):

phpstan.neon
```diff
@@ -15,9 +15,5 @@ parameters:
         -
             message: '#Parameter \#1 .* of method DoctrineORMModule\\Options\\Configuration.* stdClass given#'
             path: tests/Options/ConfigurationOptionsTest.php
-        -
```
TomHAnderson, line 18: This was removed in https://github.com/doctrine/DoctrineORMModule/commit/7f112e995ab42203ceff3dd8bd8c12cba4c6939d

PR #745, opened 2023-10-15
Added @template-extends to solve Psalm issues

driehle:
The issues came from a recent new release of laminas-stdlib, which introduced templates in AbstractOptions.

### DIFF
--- a/src/Options/DBALConfiguration.php
+++ b/src/Options/DBALConfiguration.php
@@ -8,6 +8,8 @@ use Laminas\Stdlib\AbstractOptions;
 
 /**
  * Configuration options for a DBAL Connection
+ *
+ * @template-extends AbstractOptions<mixed>
  */
 class DBALConfiguration extends AbstractOptions
 {
--- a/src/Options/DBALConnection.php
+++ b/src/Options/DBALConnection.php
@@ -12,6 +12,8 @@ use PDO;
 
 /**
  * DBAL Connection options
+ *
+ * @template-extends AbstractOptions<mixed>
  */
 final class DBALConnection extends AbstractOptions
 {
--- a/src/Options/EntityManager.php
+++ b/src/Options/EntityManager.php
@@ -6,6 +6,9 @@ namespace DoctrineORMModule\Options;
 
 use Laminas\Stdlib\AbstractOptions;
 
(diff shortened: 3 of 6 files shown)


src/Options/DBALConfiguration.php
```diff
@@ -8,6 +8,8 @@
 
 /**
  * Configuration options for a DBAL Connection
+ *
+ * @template-extends AbstractOptions<mixed>
```
Ocramius, line 12: can we be more precise here? 🤔

driehle: It would need to be the supertype of all options stored in the container. Not sure if that's desirable? Talking about all affected containers, most options are string, but there's also bool and some other objects. I don't really see the point to use template types for `AbstractOptions` anyways, but that's obviously not me to decide. 😇

Ocramius: If you add `...` to the array, you should be able to cover the whole supertype

driehle: Instructions unclear 🙈
Which array? 🤔

Ocramius: The template type to add here, in theory , is `array{foo: bar, ...}`

driehle: Maybe I am misunderstanding [AbstractOptions](https://github.com/laminas/laminas-stdlib/blob/3.19.x/src/AbstractOptions.php) somehow, but it's technically an `iterable<string, TValue>`. Therefore, the template `TValue` does not reflect the configuration array, as your snippet suggests, but rather is the type applicable to each option value. That's the reason why I am questioning the usefulness of the template as such.

Here, in this class, we have three options:
- `resultCache` as `string`
- `sqlLogger` as `?string`
- `types` as `array` or `mixed[]`, though this could probably easily made more specific using the syntax you suggested 

Therefore, we could use `@template-extends AbstractOptions<null|string|array>`, but since the relation between different options (i.e. keys) and their types is not preserved, I don't see much advantage in doing so. What is the benefit of being able to inject an array as `resultCache` type-wise, if that's not an allowed value?

TomHAnderson (approved):

PR #747, opened 2023-11-11
support psr/container 2

### LINKED ISSUE #752: orm3 compatibility
[Doctrine ORM 3](https://github.com/doctrine/orm/releases/tag/3.0.0) has been released. The module requires the corresponding compatibility.

pascalheidmann:
`psr/container` 1 and 2 are basically the same but with type hints/return types

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -63,7 +63,7 @@
         "laminas/laminas-paginator": "^2.12.2",
         "laminas/laminas-servicemanager": "^3.17.0",
         "laminas/laminas-stdlib": "^3.7.1",
-        "psr/container": "^1.1.2",
+        "psr/container": "^1.1.2 || ^2.0.0",
         "symfony/console": "^5.4.3 || ^6.0.3"
     },
     "require-dev": {


driehle:
LGTM, but we'll first need to the same for [DoctrineModule](https://github.com/doctrine/DoctrineModule/), otherwise this won't help. Would you mind providing a PR there as well?

TomHAnderson (approved):

snapshotpl:
Every run in actions install v1 because doesn't meet dependecy requirements (e.g. laminas-servicemanager)

TomHAnderson:
DoctrineModule PR https://github.com/doctrine/DoctrineModule/pull/829

SenseException:
I haven't checked every Continuous Integration build, just one lowest and one highest, but psr/container 1 is currently the only used version.

driehle (changes requested):
It doesn't make any sense to support psr/container 2 unless we upgrade laminas-servicemanager to 4.0, as 3.x supports psr/container 1 only. For that, however, we will need to go into a new major release. I have changed the milestone accordingly.

Let's wait for laminas-servicemanager 4.0 to be published. An RC is available already.

PR #749, opened 2023-12-08
PHP 8.3 support

TomHAnderson:
This PR will require at least the following:

* 8.3 support for [doctrine/DoctrineModule](https://github.com/doctrine/DoctrineModule)
  * 8.3 support for [laminas/laminas-cache-storage-adapter-filesystem](https://github.com/laminas/laminas-cache-storage-adapter-filesystem/pull/76)

Laminas must update their libraries first to support DoctrineModule

PR #751, opened 2024-01-22
Allow php 8.3

fezfez:


fezfez:
my bad didnt see https://github.com/doctrine/DoctrineORMModule/pull/749

PR #753, opened 2024-03-01
[wip] Support for ORM 3

TomHAnderson:
Composer resolves to highest versions (except phpunit)

### DIFF
--- a/tests/Assets/Entity/FormEntity.php
+++ b/tests/Assets/Entity/FormEntity.php
@@ -62,9 +62,15 @@ class FormEntity
     #[ORM\Column(type: 'string', nullable: true)]
     protected string|null $stringNullable = null;
 
+    /**
+     * This join is odd because the targetEntity exists only as an interface
+     **/
     #[ORM\OneToOne(targetEntity: Target::class)]
     protected Target $targetOne;
 
+    /**
+     * This join is odd because the targetEntity exists only as an interface
+     */
     #[ORM\OneToOne(targetEntity: Target::class)]
     #[ORM\JoinColumn(nullable: true)]
     protected Target|null $targetOneNullable = null;
@@ -85,6 +91,13 @@ class FormEntity
     protected int $specificType;
 
     /**
+     * @Form\Options({"label":"Please Choose", "value_options":{"f":"false","t":"true"}})
+     * @Form\Attributes({"type":"textarea"})
+     */
+    #[ORM\Column(type: 'integer')]
+    protected int $specificAttributeType;
+
+    /**
      * @Form\Type("DoctrineORMModule\Form\Element\EntityMultiCheckbox")
      *
      * @var FormEntityTarget[]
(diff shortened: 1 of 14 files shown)


tests/Assets/Entity/FormEntity.php
```diff
@@ -9,119 +9,105 @@
...
     protected bool $boolean;
 
-    /** @ORM\Column(type="float") */
+    #[ORM\Column(type: 'float', nullable: false)]
     protected float $float;
 
-    /** @ORM\Column(type="bigint") */
-    protected int $bigint;
+    #[ORM\Column(type: 'bigint', nullable: false)]
+    protected string $bigint;
```
TomHAnderson, line 30: old code was incorrectly cast to int

demiankatz:
@TomHAnderson, I see that this hasn't moved in a while. I'm likely to need it sooner or later. Is there anything I can do to help things along?

[edit: I've followed the trail a little further, and I see that you're running into some of the same interlocking Laminas dependency changes that have been causing trouble on my end as well... we may just have to keep waiting for some things, but if you need Laminas-related help to make progress, feel free to call on me and I'll do what I can as time permits.]

driehle:
@demiankatz We cannot provide support for ORM 3 in a new minor release. The reason is stated in doctrine/DoctrineModule#833. We need to switch DoctrineModule from `doctrine/cache` to `psr/cache`, which results in a BC break. I have, however, prioritized support for PHP 8.4, which will be available with 6.3.0, as soon as laminas/laminas-serializer#73 is merged and released.

demiankatz:
Thanks, @driehle, I definitely understand that a major release will be needed for ORM 3. I appreciate the incremental progress on minor releases in the meantime, but I'm also willing to help out as needed to get to the next major version (though I also realize that there's some uncertainty about the future of Laminas modules in general; maybe what will ultimately be needed is something of an entirely different shape... I'm still watching and waiting to see how everything ends up developing).

darckking:
Hi there.
Does anyone know if there will be a major release with ORM 3 support in observable future ?
Recently I learned about decision to eventually abandon Laminas MVC and related packages which is core part of my apps together with Laminas Doctrine ORM & ODM modules.
So I wonder if should I consider ditching Laminas framework as such in favor of alternatives.
Any advice would be appreciated.

demiankatz:
@darckking, I think a lot of people are going to be asking these questions.

I also work on a project (https://github.com/vufind-org/vufind) built on Laminas MVC (and other Laminas components). I don't think the entire Laminas ecosystem is going to go away, and certain parts like the ServiceManager remain well-supported and useful... but we definitely have to figure out how to move away from MVC, and I suspect that some related components like the ModuleManager may be abandoned (or at least significantly reduced in scope) in the future. All of that raises major questions for the future of this particular project.

Finding the best future path away from laminas-mvc will be a major topic of conversation in my project's community over the next few months. If you (or anyone else) would like to listen in on those conversations, you're more than welcome to drop in on our [community call](https://vufind.org/wiki/community_call) -- I'm always open to cross-project collaboration and discussion, and it could be mutually beneficial.

If you're looking for a method of integrating Doctrine with the Laminas ServiceManager without the entire MVC framework, you might want to take a look at https://github.com/Roave/psr-container-doctrine -- I am using this in a [much smaller, lighter weight project](https://github.com/FalveyLibraryTechnology/VuOwma). I suspect it has fewer capabilities than this project, but it might be a starting point for disentangling dependencies.

Like you, I'm just at the beginning of this journey, and I have more to learn than I presently know -- but I hope this is of some help, and I'll be happy to keep talking as we figure things out.

demiankatz:
One other thought that might be helpful: there are some pieces in this module that apply to different parts of the Laminas ecosystem -- e.g. components for Laminas\Form and an adapter for Laminas\Paginator. Would it be possible to break these out into separate projects that become dependencies of this one? That would make them more future-proof, since we could continue to use them in concert with the appropriate Laminas components even if other pieces of this module become obsolete.

demiankatz:
An addendum to the above suggestion: there's also the fact that this module depends on https://github.com/doctrine/DoctrineModule, which also has a lot of classes that support specific Laminas components. Maybe we need to start there; would it make sense, for example, to have a doctrine/LaminasFormComponents repo with all of the Form-related components from DoctrineModule and DoctrineORMModule?

I'm sure these modules were initially created when Laminas was still Zend Framework, and everything was monolithic. Now that Laminas components are much more independent, I really think it makes sense to contemplate splitting things up. This allows us to continue maintaining things that are maintainable, and it also makes it easier to abandon things that are not.

If others like this proposal and could use help from me (or the VuFind community more broadly), please let me know.

PR #754, opened 2024-04-02
Add laminas-servicemanager v4 support

samsonasik:


samsonasik:
another dependency seems use laminas-servicemanager v3, as run on php 8.3 got:

```
  - Installing laminas/laminas-servicemanager (3.22.1): Extracting archive
 ```

samsonasik:
I see, `doctrine/doctrine-module` still require v3 only.

https://github.com/doctrine/DoctrineModule/blob/905d493770193f14f19b60998de9993a3dfde5f9/composer.json#L58

samsonasik:
Ref https://github.com/doctrine/DoctrineModule/pull/835

TomHAnderson:
That's correct.  A large refactor of this project is in order but we need to wait for Laminas to catch up their own libraries before we can continue.  e.g. we need version 4 of https://github.com/laminas/laminas-cache

dkmuir:
@TomHAnderson Couldn't DoctrineORMModule be decoupled from laminas-cache? Especially since any PSR-6 adapter will do now? We migrated to Symfony's cache a few months ago as it was looking at the time that laminas-cache going stagnant and couldn't be readily relied upon long term.

TomHAnderson:
@dkmuir thanks for the news.  The Laminas, and ZF2 library before, pull from the resources available within the framework.  In that case, it would be `laminas/laminas-cache`.  It is news to me that this project is dying.  I am for using PSR standards as applicable.  But does that mean we should code against PSR-11 instead of `laminas/laminas-servicemanager`?   

I'll consult with the team.

dkmuir:
@TomHAnderson I don't think it actually is dying. I just had a look through the last year's worth of steering committee minutes, and can't find anything... Really not sure how I got that in my head... 😬

With the question of decoupling ServiceManager... I don't think it makes much sense. For those who need it, wouldn't https://github.com/Roave/psr-container-doctrine already fit the bill?

samsonasik:
laminas-cache already v4 which support laminas-servicemanager v4 https://github.com/laminas/laminas-cache/releases/tag/4.0.4

SenseException:
The checks have passed but the logs are already gone. I'm closing and reopening this PR to trigger the workflows again. I want to be sure that some checks with the "highest" setting are installing 4.0 and "lowest" 3.17.

SenseException (commented):
Looks like 3.22.1 is the highest version we get for laminas/laminas-servicemanager. Some other packages are preventing the installation of 4.x

TomHAnderson:
4.x is blocked by many Laminas modules.  Detailed here:  https://github.com/doctrine/DoctrineModule/issues/834

I've been planning on a significant upgrades to the Laminas modules but the pace of development at Laminas is slow and getting an installable set of libraries to do this upgrade is not possible at this time.

PR #769, opened 2025-01-23
Allow PHP 8.4

### LINKED ISSUE #767: Add support for PHP 8.4 to version 6
#### What

Support for PHP 8.4 should be added to version 6 of this package

#### Why

PHP 8.4 is now publicly available.  
Lack of PHP 8.4 support prevents Laminas based projects using this component from upgrading to PHP 8.4

#### How

- Update composer.json to add PHP 8.4 in the platform requirements.
- Update other dependencies on Laminas components as some may not be required or included in other dependenccies. For example, laminas-mvc includes already many dependencies that may not need to be explicitely defined.
- This can only be achieved once all the Laminas dependencies have also added support for PHP 8.4 (in progress)

driehle:
Closes #767

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -25,4 +25,4 @@ jobs:
     name: "Coding Standards"
     uses: "doctrine/.github/.github/workflows/coding-standards.yml@6.0.0"
     with:
-      php-version: "8.3"
+      php-version: "8.4"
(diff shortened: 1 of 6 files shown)


FabianKoestring:
Is there a schedule for this PR/Milestone?

driehle:
@FabianKoestring See the related ticket. We'll need laminas/laminas-serializer#73, otherwise you won't be able to install DoctrineORMModule in PHP 8.4 without using `--ignore-platform-req=php` (and, in contrast, if you were using ignore-platform-req, you could aready use DoctrineORMModule with PHP 8.4 right now).

FabianKoestring:
https://github.com/laminas/laminas-serializer/releases/tag/2.18.0 👌

driehle:
Removed again `--ignore-platform-req=php` for PHP 8.4, CI passes now as expected with the new release of laminas-serializer.

